PR #8, opened 2023-04-28
Crncc 1176 mock local services

paddyduff31:


src/StudyApi/DependencyRegistrations/InfrastructureRegistration.cs
```diff
@@ -85,6 +87,12 @@ public static IServiceCollection AddInfrastructure(this IServiceCollection servi
             if (!ProdEnvironmentNames.Any(x => string.Equals(x, environmentName, StringComparison.OrdinalIgnoreCase)))
```
con-nihr, line 87: I would change this config switch, any deployed environment won't be able to use these stubs not just prod.

Probably best to have an explicit config option to turn these on/off, you still might want to connect to the remote services locally too.

src/Infrastructure/Services/MockEmailService.cs
```diff
@@ -0,0 +1,16 @@
...
+using Application.Contracts;
+
+namespace Infrastructure.Services
+{
+    public class MockEmailService : IEmailService
+    {
+        public Task SendEmailAsync(string to, string subject, string body)
+        {
+            // log the email instead
+            Console.WriteLine($"Email sent to {to} with subject {subject} and body {body}");
```
con-nihr, line 12: We might want to output this elsewhere.
The logger, Debug console, filesystem.

Something for discussion.

con-nihr (changes requested):

src/StudyApi/DependencyRegistrations/InfrastructureRegistration.cs
```diff
@@ -81,10 +84,15 @@ public static IServiceCollection AddInfrastructure(this IServiceCollection servi
             services.AddHttpClientWithRetry<IReferenceDataApiClient, ReferenceDataApiClient>(clientsSettings.ReferenceDataService, 2, logger);
             services.AddHttpClientWithRetry<IParticipantApiClient, ParticipantApiClient>(clientsSettings.ParticipantService, 2, logger);
 
+            // get DevSwitches from appsettings.json
+            var devSwitches = configuration.GetSection(DevSwitches.SectionName).Get<DevSwitches>();
+
             // If not Prod, then enable stubs
-            if (!ProdEnvironmentNames.Any(x => string.Equals(x, environmentName, StringComparison.OrdinalIgnoreCase)))
+            if (devSwitches.EnableStubs)
```
con-nihr, line 91: I would recommend having an additional environment check to ensure this doesn't leak into production.

`if (devSwitches.EnableStubs && !ProdEnvironmentNames.Any(x => string.Equals(x, environmentName, StringComparison.OrdinalIgnoreCase)))`

con-nihr (changes requested):
Looks great, a few minor suggestions.

con-nihr (approved):

PR #10, opened 2023-05-17
Merge participant api

paddyduff31:


src/StudyApi/Extensions/EndpointRouteBuilderExtensions.cs
```diff
@@ -10,77 +9,67 @@
...
+    public static IEndpointRouteBuilder MapCustomHealthCheck(
+        this IEndpointRouteBuilder endpoints,
+        string pattern = "/api/health",
+        string servicesPattern = "/api/health/ready")
     {
-        public static IEndpointRouteBuilder MapCustomHealthCheck(
-            this IEndpointRouteBuilder endpoints,
-            string pattern = "/api/health",
-            string servicesPattern = "/api/health/ready")
+        if (endpoints == null) throw new ArgumentNullException(nameof(endpoints));
```
con-nihr, line 21: Generally you want to keep braces even for single line statements. Missing braces can lead to scope confusion and subtle bugs.

src/StudyApi/Extensions/EndpointRouteBuilderExtensions.cs
```diff
@@ -10,77 +9,67 @@
...
-                    }
-                    
-                    foreach (var (dataKey, dataValue) in value.Data)
-                    {
-                        writer.WriteString(dataKey, dataValue.ToString());
-                    }
-                    
-                    writer.WriteEndObject();
-                }
+                if (!string.IsNullOrWhiteSpace(value.Description)) writer.WriteString("description", value.Description);
```
con-nihr, line 60: Revert this change, missing braces

src/StudyApi/Extensions/EndpointRouteBuilderExtensions.cs
```diff
@@ -10,77 +9,67 @@
...
+
+                foreach (var (dataKey, dataValue) in value.Data) writer.WriteString(dataKey, dataValue.ToString());
```
con-nihr, line 62: Revert this change, missing braces

con-nihr (changes requested):
We should try to keep PRs tight and focused around a set of changes for a single feature.

I see Rider applied a lot of automatic refactoring, sometimes these are not all advisable and it will often toggle to the opposite of what is in the code if there are two ways of doing something. Widespread code formatting has to be carefully managed as it can cause all sorts of problems with merging branches and diffing commits, troubleshooting and reviewing. We don't want to pollute the change history too much for something that doesn't add lots of functional value. Sometimes it is worth it but in this case I think that formatting across the board in this solution is unnecessary.

I would recommend splitting this PR into three:

1. Merge the participant API, functional changes only.
2. The scan request updates.
3. (optional) code formatting with no functional changes. This probably won't be needed.

src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -0,0 +1,399 @@
...
+            await RemoveParticipantData(entity);
+                    
+
+            var linkedEntity = await GetParticipantDetailsByEmail(linkedEmail);
+            if (linkedEntity == null) return;
+            await RemoveParticipantData(linkedEntity);
+        }
+        catch (Exception e)
+        {
+            _logger.LogError("Delete-error = {EMessage}", e.Message);
```
con-nihr, line 190: We need the stack trace here too. 
Always pass the exception instance to the logging method where possible.
(Applies to all exception logging)

Use this overload:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logerror?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-loggerextensions-logerror(microsoft-extensions-logging-ilogger-system-exception-system-string-system-object())

con-nihr (changes requested):
Some minor suggestions / updates around async and error handling.

PR #15, opened 2023-05-31
Stop concurrent logins

paddyduff31:
Session ID is added to the user in the DB when loggining in, it is then checked on each request via middleware.  If this sessionID does not match then their session is invalidated and they are logged out


con-nihr (commented):
Some observations but nothing critical.

PR #18, opened 2023-08-22
Add regex to the backend to prevent premium number

paddyduff31:


src/Infrastructure/Services/UserService.cs
```diff
@@ -421,6 +422,14 @@ public async Task<Response<string>> RespondToMfaChallengeAsync(string mfaCode, s
 
         private string CleanPhoneNumber(string phoneNumber)
         {
+            // ensure its a valid uk mobile using regex
+            var isValid = Regex.IsMatch(phoneNumber, @"^((\+44\s?|0)7([45789]\d{2}|624)\s?\d{3}\s?\d{3})$",
```
con-nihr, line 426: The use of a single regex here could be problematic.
I suggest using [libphonenumber-csharp](https://github.com/caseykramer/libphonenumber-csharp) to parse, validate and format the number in E164 (that's what cognito is expecting).

src/Infrastructure/Services/UserService.cs
```diff
@@ -421,6 +422,14 @@ public async Task<Response<string>> RespondToMfaChallengeAsync(string mfaCode, s
...
+                RegexOptions.None, TimeSpan.FromMilliseconds(100));
+            if (!isValid)
+            {
+                throw new ValidationException();
```
con-nihr, line 430: A message in the exception constructor would be useful here, e.g. "'xyz' is not a valid mobile phone number."

src/Infrastructure/Services/UserService.cs
```diff
@@ -433,61 +442,75 @@ private string CleanPhoneNumber(string phoneNumber)
...
-                    .Replace("###LINK_REPLACE###", "")
-                    .Replace("###LINK_DISPLAY_VALUE_REPLACE###", "block")
-                    .Replace("###TEXT_REPLACE5###",
-                        "")
-                    .Replace("###TEXT_REPLACE6###", "");
 
-                await _emailService.SendEmailAsync(email, "Be Part of Research mobile phone number verified", htmlBody);
+                await _provider.AdminUpdateUserAttributesAsync(request);
             }
+            catch (Exception e)
```
con-nihr, line 504: You should catch the most specific relevant exception.
`catch(ValidationException ex)`
in which case the status code will be BadRequest rather than InternalServerError.

Might be a good idea to create a new application specific exception to throw at the end. e.g. `CognitoPhoneNumberUpdateException`. Throwing a generic `Exception` is almost always a bad idea.

con-nihr (changes requested):

src/Infrastructure/Services/UserService.cs
paddyduff31: I have created the generic email message but still getting a 500 status code, perhaps you can advise?

Jaffacakes82 (approved):

src/Infrastructure/Services/UserService.cs
```diff
@@ -425,73 +428,115 @@ public async Task<Response<string>> RespondToMfaChallengeAsync(string mfaCode, s
...
+            if (phoneUtil.GetNumberType(number) != PhoneNumberType.MOBILE)
+            {
+                throw new ValidationException(new List<ValidationFailure>
+                {
+                    new ValidationFailure("PhoneNumber", $"{number} is not a valid UK mobile number")
+                }, "PhoneNumber");
+            }
+
+            // Ensure the phone number is in the correct format for cognito i.e., +441234567890 E164
+            phoneNumber = phoneUtil.Format(number, PhoneNumberFormat.E164);
```
con-nihr, line 453: reusing the input parameter here, just `return phoneUtil.Format(number, PhoneNumberFormat.E164);`

src/Infrastructure/Services/UserService.cs
```diff
@@ -425,73 +428,115 @@ public async Task<Response<string>> RespondToMfaChallengeAsync(string mfaCode, s
...
+            PhoneNumberUtil phoneUtil = PhoneNumberUtil.GetInstance();
+
+            // Ensure it's a valid UK number using libphonenumber
+            var number = phoneUtil.Parse(phoneNumber, "GB");
+            if (!phoneUtil.IsValidNumber(number))
             {
-                phoneNumber = $"+44{phoneNumber.TrimStart('0')}";
+                throw new ValidationException(new List<ValidationFailure>
+                {
+                    new ValidationFailure("PhoneNumber", $"{number} is not a valid UK number")
```
con-nihr, line 439: `"Phone number ...`

src/Infrastructure/Services/UserService.cs
```diff
@@ -425,73 +428,115 @@ public async Task<Response<string>> RespondToMfaChallengeAsync(string mfaCode, s
...
+                }, "PhoneNumber");
             }
 
+            // Check if the number is specifically a mobile number
+            if (phoneUtil.GetNumberType(number) != PhoneNumberType.MOBILE)
+            {
+                throw new ValidationException(new List<ValidationFailure>
+                {
+                    new ValidationFailure("PhoneNumber", $"{number} is not a valid UK mobile number")
```
con-nihr, line 448: `"Phone number ...`

PR #21, opened 2023-08-22
pass back generic error message

paddyduff31:


src/StudyApi/Common/RequestModelValidatorFilter.cs
```diff
@@ -22,7 +22,15 @@ public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionE
                 var errors = context.ModelState.Values.SelectMany(v => v.Errors.Select(b => b.ErrorMessage)).ToList();
                 var errorsString = string.Join("; ", errors);
                 _logger.LogWarning("Request Validation Failed: {Errors}", errorsString);
-                context.Result = new BadRequestObjectResult(context.ModelState);
+
+                // Set a generic error message
+                var genericErrorResponse = new
```
matthetherington-pa, line 27: just a suggestion, but the default behaviour is really useful for local dev

consider adding a check to return the generic message only when ASPNETCORE_ENVIRONMENT == Production, otherwise you get the original behaviour (in effect anyway, you'd be better adding an option in appsettings)

Do you know what the motivation for the change is? The only potential vulnerability I can think of that this would help mitigate is exploiting an issue with the JSON parsing, it seems quite low risk

matthetherington-pa (approved):

PR #24, opened 2023-09-06
Crncc 1461 vs cms welsh 1 6 emails

paddyduff31:


src/Application/Constants/ContentfulEmailNames.cs
```diff
@@ -0,0 +1,10 @@
+namespace Application.Constants
+{
+    public class ContentfulEmailNames
```
con-nihr, line 3: Are these constants deployment specific? In which case they should be read from configuration rather than compile time constants.

src/Application/Contracts/IUserService.cs
```diff
@@ -8,7 +8,7 @@ public interface IUserService
     {
         Task<Response<string>> LoginAsync(string email, string password);
         Task<Response<NhsLoginResponse>> NhsLoginAsync(string code, string redirectUrl);
-        Task<Response<SignUpResponse>> NhsSignUpAsync(bool consent, string token);
+        Task<Response<SignUpResponse>> NhsSignUpAsync(bool consent, string selectedLanguage, string token);
```
con-nihr, line 11: I would be inclined to added the new parameter at the end of the signature to make maintenance easier (unless there is a good reason to add it sooner, like importance). Especially when the types are the same, easy to get `token` and `selectedLanguage` mixed up when updating existing call sites.

I would recommend using the `CultureInfo selectedLocale` instead.

src/Application/Contracts/IContentfulService.cs
```diff
@@ -0,0 +1,10 @@
+using System.Threading.Tasks;
+using Application.Responses.V1;
+
+namespace Application.Contracts
+{
+    public interface IContentfulService
+    {
+        Task<ContentfulEmail> GetContentfulEmailAsync(string entryId, string locale);
```
con-nihr, line 8: `CultureInfo locale`

src/Application/Users/V1/Commands/NhsSignUpCommand.cs
```diff
@@ -10,11 +10,13 @@ namespace Application.Users.V1.Commands
     public class NhsSignUpCommand : IRequest<Response<SignUpResponse>>
     {
         private bool ConsentRegistration { get; }
+        private string SelectedLanguage { get; }
         private string Token { get; }
 
-        public NhsSignUpCommand(bool consentRegistration, string token)
+        public NhsSignUpCommand(bool consentRegistration, string selectedLanguage, string token)
```
con-nihr, line 16: `CultureInfo locale`

src/Application/Users/V1/Commands/NhsSignUpCommand.cs
```diff
@@ -10,11 +10,13 @@ namespace Application.Users.V1.Commands
     public class NhsSignUpCommand : IRequest<Response<SignUpResponse>>
     {
         private bool ConsentRegistration { get; }
+        private string SelectedLanguage { get; }
```
con-nihr, line 13: `CultureInfo SelectedLocale`

src/Domain/Entities/Participants/ParticipantDetails.cs
```diff
@@ -16,6 +16,7 @@ public class ParticipantDetails
         [DynamoDBProperty] public string NhsId { get; set; }
         [DynamoDBProperty] public string SessionId { get; set; }
         [DynamoDBProperty] public string NhsNumber { get; set; }
+        [DynamoDBProperty] public string SelectedLanuguage { get; set; }
```
con-nihr, line 19: Typo, but change `CultureInfo SelectedLocale` with converter if needed.

src/Infrastructure/Services/UserService.cs
```diff
@@ -183,6 +187,28 @@ public async Task<Response<NhsLoginResponse>> NhsLoginAsync(string code, string
             }
         }
 
+        private async Task SendContentfulEmailAsync(string emailName, string emailRecipient, string selectedLanguage,
```
con-nihr, line 190: `CultureInfo selectedLocale`

src/Infrastructure/Services/UserService.cs
```diff
@@ -183,6 +187,28 @@ public async Task<Response<NhsLoginResponse>> NhsLoginAsync(string code, string
...
+            var contentfulEmail = await _contentfulService.GetContentfulEmailAsync(emailName, selectedLanguage);
+            string htmlContent = _richTextToHtmlConverter.Convert(contentfulEmail.EmailBody);
+
+            if (!string.IsNullOrEmpty(firstName))
+            {
+                var template = Handlebars.Compile(htmlContent);
+                var data = new
+                {
+                    firstName =
+                        System.Threading.Thread.CurrentThread.CurrentCulture.TextInfo.ToTitleCase(firstName.ToLower()),
```
con-nihr, line 202: Use the selected locale to do text manipulation `selectedLocale.TextInfo.ToTitleCase(firstName)`

src/Infrastructure/Services/UserService.cs
```diff
@@ -193,7 +219,8 @@ private Response<NhsLoginResponse> HandleNhsLoginException(Exception ex)
             return exceptionResponse;
         }
 
-        public async Task<Response<SignUpResponse>> NhsSignUpAsync(bool consentRegistration, string token)
+        public async Task<Response<SignUpResponse>> NhsSignUpAsync(bool consentRegistration, string selectedLanguage,
```
con-nihr, line 222: `CultureInfo selectedLocale`

src/StudyApi/Controllers/V1/Users/UsersController.cs
```diff
@@ -158,6 +158,7 @@ public async Task<IActionResult> SignUpUserAsync([FromBody] SignUpRequest reques
         public class NhsSignUpRequestLocal
         {
             public bool ConsentRegistration { get; set; }
+            public string SelectedLanguage { get; set; }
```
con-nihr, line 161: `CultureInfo SelectedLocale`

src/Infrastructure/Services/ContentfulService.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class ContentfulService: IContentfulService
+    {
+        private readonly IContentfulClient _client;
+
+        public ContentfulService(IContentfulClient client)
+        {
+            _client = client;
+        }
+
+        public async Task<ContentfulEmail> GetContentfulEmailAsync(string entryId, string locale = "en-GB")
```
con-nihr, line 18: `CultureInfo locale` default can be handled with `null`.

`if(locale==null) { locale = new CultureInfo("en-GB"); }`

src/Infrastructure/Services/ContentfulService.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            var entry = await _client.GetEntry(entryId, new QueryBuilder<ContentfulEmail>().LocaleIs(locale));
```
con-nihr, line 20: `LocaleIs()` should accept `CultureInfo locale` directly without having to call `ToString()`.

con-nihr (changes requested):

src/Application/Constants/ContentfulEmailNames.cs
paddyduff31: These are not deployment specific

src/Infrastructure/Services/ContentfulService.cs
paddyduff31: Great thanks, this was my only question around using CultureInfo

src/Application/Contracts/IUserService.cs
paddyduff31: I just grouped the first 2 together as these are both "user actions".  I have updated the type to CultureInfo and now the work is done I think the risk is minimised, changing back would be unnecessary but I will keep in mind for future dev

src/Infrastructure/Services/ContentfulService.cs
paddyduff31: I had to ToString, it wouldn't accept CultureInfo

src/Infrastructure/Services/UserService.cs
```diff
@@ -183,6 +188,28 @@ public async Task<Response<NhsLoginResponse>> NhsLoginAsync(string code, string
...
+            var contentfulEmail = await _contentfulService.GetContentfulEmailAsync(emailName, selectedLocale);
+            string htmlContent = _richTextToHtmlConverter.Convert(contentfulEmail.EmailBody);
+
+            if (!string.IsNullOrEmpty(firstName))
+            {
+                var template = Handlebars.Compile(htmlContent);
+                var data = new
+                {
+                    firstName =
+                        selectedLocale.TextInfo.ToTitleCase(firstName.ToLower()),
```
con-nihr, line 203: No need to call `ToLower()` here as that part is not locale specific and we are shifting it into TitleCase anyway.

con-nihr (changes requested):

src/Application/Constants/ContentfulEmailNames.cs
paddyduff31: They are the id of the contentful entry.  I take your point though as if someone deletes the email then creates a new one it'll need fixed instead of just updating secrets

src/Infrastructure/Services/UserService.cs
paddyduff31: I tried removing the ToLower and it sent through the firstName as all Caps so I had to add it again to work

src/Infrastructure/Services/ContentfulService.cs
```diff
@@ -0,0 +1,26 @@
...
+        private readonly IContentfulClient _client;
+
+        public ContentfulService(IContentfulClient client)
+        {
+            _client = client;
+        }
+
+        public async Task<ContentfulEmail> GetContentfulEmailAsync(string entryId, CultureInfo locale)
+        {
+            if (locale==null) { locale = new CultureInfo("en-GB"); }
```
con-nihr, line 21: Remove this null check and default value, see below.

src/Infrastructure/Services/UserService.cs
```diff
@@ -183,6 +191,29 @@ public async Task<Response<NhsLoginResponse>> NhsLoginAsync(string code, string
             }
         }
 
+        private async Task SendContentfulEmailAsync(string emailName, string emailRecipient, CultureInfo selectedLocale,
+            string firstName = null)
+        {
+            if (selectedLocale == null) selectedLocale = new CultureInfo("en-GB");
```
con-nihr, line 197: Generally you need to braces here to make it explicit which statements are covered by the `if`.
While leaving out the braces is valid code, it is the source of quite a few bugs.

`if (selectedLocale == null) { selectedLocale = new CultureInfo("en-GB");}`
or even better:
`selectedLocale = selectedLocale ?? new CultureInfo("en-GB");`

However, I see now that we need to set the default locale in a few different places which could be a problem.
Best to remove the check in `ContentfulService.cs` and set it once as early in the control flow as possible. Looks like here is the best place.

src/Infrastructure/Services/UserService.cs
con-nihr: Is it coming through originally in all caps? That might be why, it is being interpreted as an abbreviation.

In that case the `ToLower()` has to be locale aware.

`selectedLocale.TextInfo.ToTitleCase(firstName.ToLower(selectedLocale)))`

con-nihr (changes requested):

src/Infrastructure/Services/UserService.cs
paddyduff31: Yes all caps.  I will update

paddyduff31: why does the `ToLower()` need to be locale aware?  And the title case for that matter?

con-nihr: In isolation the characters are just unicode symbols, they don't mean anything except when put into the context of a locale. Different locales will have different rules about what symbols are uppercase/lowercase and how you map between them, it might even be specific to where the character appears in the word or when it is next to a certain character.

So to make something lower case, you need to know that it is lowercase in English or lowercase in French (maybe there are rules for accents, etc.)

In this PR we want to change the case of the first name and rather than use the culture rules for the server we should use the culture rules of the email that we are constructing.

Some further reading:
https://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/

That is also why when you want to make a conversion or comparison that always behaves the same regardless of what the user or server locale settings are you use the "invariant culture". This is useful for internal calculations and things you want to be consistent (even if they are technically wrong - at least it is not a weird bug that only happens when the server has it's locale set to en-US).

Anything that is user facing like these emails should be fully culture aware. Anything internal that some functionality depends on should default to using the InvariantCulture unless there is a good reason to do otherwise.

https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=net-7.0

con-nihr (approved):

con-nihr:
Approved, though you may want to consolidate some of these changes with the event handler PR.

PR #30, opened 2023-09-29
initial commit

paddyduff31:


src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -94,19 +94,32 @@ private async Task<string> AdminGetUserAsync(string email)
...
         }
 
+        // get the linked account row by passing the ParticipantId into the pk as the new row saves the PK as ParticipantId to link 
+        var linkedRecord =
+            await _participantRepository.GetParticipantAsync(
+                $"PARTICIPANT#{oldUser.ParticipantId}");
+
+        // delete the linked row
+        await _participantRepository.DeleteParticipantAsync(linkedRecord);
+
```
con-nihr, line 117: Consider what the worst case scenario is in this case.

How 'bad' would it be if the code fails here?
The old record has been deleted before the new linked record is created.
Is that recoverable?

Would it be better to create the new record before deleting the old one?
Is that better or worse if the process fails between those two operations?

src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -181,25 +194,25 @@ public async Task DeleteUserAsync(string participantId)
...
-                    "")
-                .Replace("###LINK_REPLACE###", "")
-                .Replace("###LINK_DISPLAY_VALUE_REPLACE###", "block")
-                .Replace("###TEXT_REPLACE5###",
-                    "")
-                .Replace("###TEXT_REPLACE6###",
-                    "");
-
-            await _emailService.SendEmailAsync(entity.Email, "Be Part of Research", htmlBody);
+            // var baseUrl = _emailSettings.WebAppBaseUrl;
```
con-nihr, line 197: Delete these commented out lines. They are already in source control if you need to recover them.

src/Application/Contracts/IParticipantRepository.cs
```diff
@@ -9,12 +12,20 @@ public interface IParticipantRepository
...
         Task UpdateParticipantDetailsAsync(ParticipantDetails entity);
         Task CreateParticipantDemographicsAsync(ParticipantDemographics entity);
         Task AddDemographicsToNhsUserAsync(ParticipantDemographics entity, string nhsId);
         Task UpdateParticipantDemographicsAsync(ParticipantDemographics entity);
         Task DeleteParticipantDetailsAsync(ParticipantDetails entity);
+        
         Task CreateAnonymisedDemographicParticipantDataAsync(ParticipantDetails entity);
         Task<ParticipantDetails> QueryIndexForParticipantDetailsAsync(string query, string colName);
+        Task CreateParticipantAsync(Participant entity);
+        IAsyncEnumerable<Participant> GetAllAsync(
```
con-nihr, line 25: Make `IParticipantRepository` implement `IAsyncEnumerable<Participant>`.
`IParticipantRepository : IAsyncEnumerable<Participant> { }`

The method `GetAllAsync()` is confusing and unnecessary.
You want to enumerate the repository, so you can just do that directly. Also the `...Async` suffix makes it look like you have to await this call directly.

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <response code="200">When IsSuccess true</response>
+    /// <response code="500">Server side error</response>
+    [AllowAnonymous]
+    [SwaggerResponse(StatusCodes.Status200OK, Type = typeof(Response<UserLoginResponse>))]
+    [SwaggerResponse(StatusCodes.Status500InternalServerError, Type = null)]
+    [HttpPost("UpdateLinkedRecords")]
+    public async Task<IActionResult> UpdateLinkedRecords()
+    {
+        var userRecordErrors = new List<UserRecordError>();
+        var userRecords = _participantRepository.GetAllAsync();
```
con-nihr, line 49: No longer needed.

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+        await foreach (var userRecord in userRecords)
```
con-nihr, line 50: `IParticipantRepository` now implements `IAsyncEnumerable<Participant>` so this becomes:
`await foreach (var userRecord in _participantRepository)`

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <summary>
+    /// [AllowAnonymous] Login
+    /// </summary>
+    /// <response code="200">When IsSuccess true</response>
+    /// <response code="500">Server side error</response>
+    [AllowAnonymous]
+    [SwaggerResponse(StatusCodes.Status200OK, Type = typeof(Response<UserLoginResponse>))]
+    [SwaggerResponse(StatusCodes.Status500InternalServerError, Type = null)]
+    [HttpPost("UpdateLinkedRecords")]
+    public async Task<IActionResult> UpdateLinkedRecords()
```
con-nihr, line 46: I know this is a local harness but consider capturing the request `CancellationToken` here and passing it down through the layers.

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+                catch (Exception e)
+                {
+                    // store errors
+                    userRecordErrors.Add(new UserRecordError
+                    {
+                        UserRecord = userRecord,
+                        Error = e.Message
+                    });
```
con-nihr, line 87: If this is for development use only, might as well capture the whole exception. Add a property `Exception = e`.

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    await _participantService.DeleteUserAsync(userRecord.ParticipantId);
+                }
+            }
+            catch (Exception e)
+            {
+                userRecordErrors.Add(new UserRecordError
+                {
+                    UserRecord = userRecord,
+                    Error = e.Message
```
con-nihr, line 137: Capture exception.

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+                });
+            }
+        }
+
+        Console.WriteLine(userRecordErrors);
+
+        return Ok(userRecordErrors);
```
con-nihr, line 144:
```
return userRecordErrors.Any()
            ? StatusCode(500, userRecordErrors)
            : Ok();
```

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                userRecordErrors.Add(new UserRecordError
+                {
+                    UserRecord = userRecord,
+                    Error = e.Message
+                });
+            }
+        }
+
+        Console.WriteLine(userRecordErrors);
```
con-nihr, line 142: `Debug.WriteLine` to output to the development console rather than the host.

con-nihr (changes requested):

src/Application/Contracts/IParticipantRepository.cs
paddyduff31: I have done this and it made me implement another method.  Still not sure how this is working, might be worth a catch up?  I understand WHAT it is doing just not the how

con-nihr: `IAsyncEnumerable` is a 'built in' interface that the compiler looks for when iterating over a collection with `await foreach`.

Under the hood it calls the implementation of `IAsyncEnumerable.GetAsyncEnumerator()` on your class and iterates over that.

That will be the method you were forced to implement and you would just put your implementation of `GetAllAsync()` in there.

Doing it this way makes the intent/behaviour of your class more evident. It says "`IParticipantRepository` is something you can iterate over asynchronously because it implements the `IAsyncEnumerable` interface."

`GetAllAsync()` is something custom to this codebase and is not based on the language conventions so a reader can't infer too much about what it does right away.

In fact it is slightly confusing as it might appear that it should be used like this:
`List<Participant> participants = await _participantRepository.GetAllAsync();`

See:
https://learn.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8#a-tour-through-async-enumerables

src/Infrastructure/Services/ParticipantService.cs
paddyduff31: This is for reporting purposes only so nothing really application critical.  I have moved to the end of the file though as there is more important logic before this.  I have also moved the creation before the deletion.

src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
```diff
@@ -68,7 +73,7 @@ public async Task<ParticipantDetails> QueryIndexForParticipantDetailsAsync(strin
                 _logger.LogInformation("item: {Item}", JsonConvert.SerializeObject(item, Formatting.Indented));
 
                 var participant = _context.FromDocument<ParticipantDetails>(Document.FromAttributeMap(item));
-                return await GetParticipantDetailsAsync(participant.Pk.Replace("PARTICIPANT#", ""));
+                return await GetParticipantDetailsAsync(participant.Pk.Replace("PARTICIPANT#", ""), cancellationToken);
             }
             catch (Exception e)
```
con-nihr, line 78: No need for this `try-catch` if you are going to output and rethrow.
Your debugger should break when the exception is first thrown.

src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
```diff
@@ -77,57 +82,105 @@ public async Task<ParticipantDetails> QueryIndexForParticipantDetailsAsync(strin
...
 
-        public async Task CreateAnonymisedDemographicParticipantDataAsync(ParticipantDetails entity)
+        public async Task CreateAnonymisedDemographicParticipantDataAsync(ParticipantDetails entity,
+            CancellationToken cancellationToken = default)
         {
-            await _context.SaveAsync(entity, _config);
+            await _context.SaveAsync(entity, _config, cancellationToken);
+        }
+
+        public async IAsyncEnumerable<Participant> GetAllAsync(
```
con-nihr, line 146: Remove this method and put the implementation in `GetAsyncEnumerator()`

src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
```diff
@@ -77,57 +82,105 @@ public async Task<ParticipantDetails> QueryIndexForParticipantDetailsAsync(strin
...
+        }
+
+        public async Task CreateParticipantAsync(Participant entity, CancellationToken cancellationToken = default)
+        {
+            await _context.SaveAsync(entity, _config, cancellationToken);
+        }
+
+        public async IAsyncEnumerator<Participant> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+        {
+            await foreach (var participant in GetAllAsync(cancellationToken))
```
con-nihr, line 180: You are iterating this enumerator and wrapping it in another async/await.

Move the implementation of `GetAllAsync()` in here.

con-nihr (changes requested):

src/StudyApi/Controllers/V1/Local/LocalController.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// <response code="200">When IsSuccess true</response>
+    /// <response code="500">Server side error</response>
+    [AllowAnonymous]
+    [SwaggerResponse(StatusCodes.Status200OK, Type = typeof(Response<UserLoginResponse>))]
+    [SwaggerResponse(StatusCodes.Status500InternalServerError, Type = null)]
+    [HttpPost("UpdateLinkedRecords")]
+    public async Task<IActionResult> UpdateLinkedRecords(CancellationToken cancellationToken = default)
+    {
+        var userRecordErrors = new List<UserRecordError>();
+        await foreach (var userRecord in _participantRepository)
```
con-nihr, line 49: Add this before merging:
`await foreach (var userRecord in _participantRepository.WithCancellation(cancellationToken))`

con-nihr (approved):

PR #36, opened 2023-12-14
Convert ParticipantIdentifier.Value to Guid. Add index.

con-nihr:
Simplify local dev configuration. Use appsettings.user.json. Robustly handle update / delete when record does not exist in target db.


Dynamo.Stream.Handler/src/Dynamo.Stream.Handler/Entities/ParticipantDbContext.cs
```diff
@@ -38,34 +37,25 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+    public static IQueryable<Participant> ForUpdate(this IQueryable<Participant> source)
     {
-        pk = StripPrimaryKey(pk);
-        return await Participants
-            .Where(x => x.ParticipantIdentifiers.Any(y => y.Value == pk))
-            .Include(x => x.Address)
-            .Include(x => x.HealthConditions)
-            .Include(x => x.ParticipantIdentifiers)
-            .SingleOrDefaultAsync(cancellationToken);
+        return source.Include(x => x.Address)
```
paddyduff31, line 57: This fails, lets discuss

paddyduff31 (approved):

PR #39, opened 2023-12-28
handle error in controller and pass successful response

paddyduff31:


src/Infrastructure/Services/UserService.cs
```diff
@@ -916,8 +916,7 @@ public async Task<Response<SignUpResponse>> SignUpAsync(string email, string pas
                         await _emailService.SendEmailAsync(email, "Be Part of Research registration attempt", htmlBody);
                     }
 
-                    return Response<SignUpResponse>.CreateSuccessfulContentResponse(
-                        new SignUpResponse { IsSuccess = true, }, _headerService.GetConversationId());
+                    throw new UsernameExistsException("User already exists");
```
con-nihr, line 919: Make the messages consistent. As this is already a `UsernameExistsException` just passe the email address as the message.

src/Infrastructure/Services/UserService.cs
```diff
@@ -943,8 +942,7 @@ public async Task<Response<SignUpResponse>> SignUpAsync(string email, string pas
 
                     await _emailService.SendEmailAsync(email, "Be Part of Research registration attempt", htmlBody);
 
-                    return Response<SignUpResponse>.CreateSuccessfulContentResponse(
-                        new SignUpResponse { IsSuccess = true, }, _headerService.GetConversationId());
+                    throw new UsernameExistsException("Username already exists");
```
con-nihr, line 945: Make the messages consistent. As this is already a `UsernameExistsException` just passe the email address as the message.

con-nihr (changes requested):

paddyduff31:
reported by louise as part of CMS work so dealing with there

PR #50, opened 2024-01-26
add dev setting to check for mfa bypass

paddyduff31:


src/Infrastructure/Services/UserService.cs
```diff
@@ -399,6 +404,13 @@ public async Task<Response<string>> RespondToMfaChallengeAsync(string mfaCode, s
             try
             {
                 var mfaLoginDetails = DeserializeMfaLoginDetails(mfaDetails);
+                if (_devSettings.CurrentValue.BypassMfa)
```
Jaffacakes82, line 407: Do we set ASPNETCORE_ENVIRONMENT in our environment deployments? Perhaps we should.. and then add that as an extra layer of protection here, as this could easily be misconfigured in PROD.

paddyduff31: currently being set to Production for all envs, I agree that this would be a sensible addition but may require a bigger re-write due to secrets and when they are configured based off env but I will take a look Monday

src/StudyApi/Common/RequestModelValidatorFilter.cs
```diff
@@ -28,7 +28,7 @@ public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionE
                 _logger.LogWarning("Request Validation Failed: {Errors}", errorsString);
 
                 // Set a generic error message if not in development
-                if (!_environment.IsDevelopment())
+                if (!_environment.IsEnvironment("local") || !_environment.IsDevelopment())
```
con-nihr, line 31: We should avoid sprinkling these environment checks throughout the code and setup the environment with dependency injection.

src/StudyApi/Extensions/AwsSecretsConfigurationBuilderExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static IWebHostBuilder AddAwsSecrets(this IWebHostBuilder hostBuilder)
         {
             return hostBuilder.ConfigureAppConfiguration((hostingContext, configBuilder) =>
             {
-                if (hostingContext.HostingEnvironment.IsDevelopment()) return;
+                if (!hostingContext.HostingEnvironment.IsEnvironment("local")) return;
```
con-nihr, line 20: `local` = `Development` under .Net conventions.

We should have a shared AWS deployed environment configuration extension that is called in `Staging` and `Production`.

con-nihr (changes requested):
There are some conventions that we'll want to keep in mind to avoid giving ourselves more work.

.Net by default assumes local development is the `Development` environment, the local IDE and project system uses this out of the box.

We clash with this by having a deployed `Development` environment which .Net would consider one of the `Staging` environments.

To align with .Net conventions:
`Development` = local
`Staging` = dev, test, uat, oat (non production environments)
`Production` = production

We can, of course, deviate from this but we have to work against things that happen automatically.

I think we should separate these two concepts and use `ASPNETCORE_ENVIRONMENT` in line with the .Net conventions and have a separate variable e.g. `ENVIRONMENT_LABEL` to hold our development stage label/name if needed.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-8.0

paddyduff31:
I have moved the service registration into the infra registration for more flexibility so this should fail in production which will stop it bleeding in.  Happy to catch up tomorrow @con-nihr if there is an alternative

src/Infrastructure/Services/UserService.cs
```diff
@@ -70,11 +77,13 @@ public class UserService : IUserService
...
+            _devSettings = devSettings;
             _participantService = participantService;
             _dataProtector = dataProtector.CreateProtector("mfa.login.details");
             _contentfulService = contentfulService;
             _contentfulSettings = contentfulSettings;
+            _mockIdentityService = mockIdentityService;
+            _environment = environment;
         }
 
         private string GenerateMfaDetails(AdminInitiateAuthResponse response, string password = null)
```
con-nihr, line 89: This is the method that has different behaviour depending on settings / environment.

This should be a separate interface with production / development implementations configured on startup.

src/StudyApi/DependencyRegistrations/InfrastructureRegistration.cs
```diff
@@ -91,17 +97,22 @@ public static class InfrastructureRegistration
                 clientsSettings.ReferenceDataService, 2, logger);
 
             var devSettings = configuration.GetSection(DevSettings.SectionName).Get<DevSettings>();
+            services.Configure<DevSettings>(configuration.GetSection(DevSettings.SectionName));
 
             // If not Prod, then enable stubs
-            if (devSettings.EnableStubs && !ProdEnvironmentNames.Any(x =>
-                    string.Equals(x, environmentName, StringComparison.OrdinalIgnoreCase)))
+            if (devSettings.EnableStubs && IsProdEnvironment(environmentName))
```
con-nihr, line 103: `!IsProdEnvironment()`

src/StudyApi/DependencyRegistrations/InfrastructureRegistration.cs
```diff
@@ -91,17 +97,22 @@ public static class InfrastructureRegistration
...
             {
                 // Enable local stubs
                 services.AddScoped<IEmailService, MockEmailService>();
                 services.AddSingleton<IAmazonCognitoIdentityProvider, MockCognitoProvider>();
             }
 
+            if (!IsProdEnvironment(environmentName))
+            {
+                services.AddTransient<IMockIdentityService, MockIdentityService>();
```
con-nihr, line 112: Move and combine above configuration into this `!IsProdEnvironment` block.

src/StudyApi/DependencyRegistrations/InfrastructureRegistration.cs
```diff
@@ -30,6 +31,12 @@ public static class InfrastructureRegistration
     {
         private static readonly string[] ProdEnvironmentNames = { "production", "prod", "live" };
 
+        private static bool IsProdEnvironment(string environmentName)
```
con-nihr, line 34: Do we need to be this tolerant of the environment name? The configuration should be safe for any non-production environment and production settings are used only for `production`.

If we do need this...

Use extension method similar to how the framework implements things.
https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/HostEnvironmentEnvExtensions.cs#L42

Also use the built in `IsEnvironment(string)` check...

`public static bool HasProductionLikeName(this IHostEnvironment env) => ProdEnvironmentNames.Any(x => env.IsEnvironment(x));`

con-nihr (changes requested):

src/StudyApi/DependencyRegistrations/InfrastructureRegistration.cs
paddyduff31: I was just following what is currently in play, I have updated to use the environment directly with the extension method

paddyduff31:
@con-nihr in this line:

`services.Decorate<IAuthenticationService, DevAuthenticationService>();`

Surely this could just be 

`services.AddTransient<IAuthenticationService, DevAuthenticationService>();`

Can you explain or is this just a convention thing?  Or have I implemented wrong?

src/Application/Contracts/IUserService.cs
```diff
@@ -38,5 +42,10 @@ public interface IUserService
         Task<Response<string>> ValidateEmailOtpAsync(string requestMfaDetails, string code);
         Task<string> GetMaskedMobile(string requestMfaDetails);
         Task<Response<string>> ReissueMfaSessionAsync(string requestMfaDetails);
+        Response<string> HandleMfaException(Exception ex, string errorType);
+        AdminRespondToAuthChallengeRequest CreateAuthChallengeRequest(string challengeName, string sessionId,
+            string username, string code, string codeKey);
+        MfaLoginDetails DeserializeMfaLoginDetails(string mfaDetails);
```
con-nihr, line 48: I don't think this makes sense on this interface.
The implementation is very simple so you could duplicate the code if you wanted to or add it to `MfaLoginDetails`.

src/Application/Models/MFA/MfaLoginDetails.cs
```diff
@@ -0,0 +1,9 @@
+namespace Application.Models.MFA;
+
+public class MfaLoginDetails
+{
+    public string Username { get; set; }
+    public string SessionId { get; set; }
+    public string Password { get; set; }
+    public string PhoneNumber { get; set; }
+}
```
con-nihr, line 9: Could add 
`public static MfaLoginDetails FromProtectedString(IDataProtectionProvider dataProtector, string value) => JsonConvert.DeserializeObject<MfaLoginDetails>(dataProtector.Unprotect(value));`

src/Infrastructure/Decorators/DevAuthenticationService.cs
con-nihr, line 1: Rather than `Decorators` put this in `Infrastructure/Services/Development` with other development only services.

src/Infrastructure/Decorators/DevAuthenticationService.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if (_devSettings.CurrentValue.BypassMfa)
+        {
+            var mfaLoginDetails = _userService.DeserializeMfaLoginDetails(mfaDetails);
+            return Response<string>.CreateSuccessfulContentResponse(
+                CreateLocalIdToken(mfaLoginDetails.Username),
+                _headerService.GetConversationId());
+        }
+
+        return await _authenticationService.RespondToMfaChallengeAsync(mfaCode, mfaDetails);
```
con-nihr, line 96: Put this in an else block because only one or other of these should happen.

This is distinct from `SignUpAsync()` where the inner service must be called in all cases.

src/Infrastructure/Services/MockEmailService.cs
con-nihr, line 1: Move this to `Infrastructure/Services/Development`. Rename to `NullEmailService`.
A Null implementation is the default, do nothing behaviour.

See https://sourcemaking.com/design_patterns/null_object

src/Infrastructure/Services/UserService.cs
```diff
@@ -175,11 +172,11 @@ public async Task<Response<string>> LoginAsync(string email, string password)
             }
         }
 
-        private MfaLoginDetails DeserializeMfaLoginDetails(string mfaDetails) =>
+        public MfaLoginDetails DeserializeMfaLoginDetails(string mfaDetails) =>
```
con-nihr, line 175: Move to `MfaLoginDetails`

con-nihr (changes requested):

src/Infrastructure/Decorators/DevAuthenticationService.cs
paddyduff31: Surely the else block is redundant due to the return keyword?

con-nihr: Functionally redundant yes, but not in terms of making the intent of the code explicit and more robust.

If someone changes the if block on line 88 so it doesn't return, then you fall through to call to the inner service which may be problematic. With an else block you will get a compiler error (which is a good thing).

con-nihr (changes requested):

con-nihr (approved):

PR #54, opened 2024-03-05
ability to locally delete accounts

paddyduff31:


src/Infrastructure/Services/Development/NullEmailService.cs
```diff
@@ -1,16 +1,36 @@
...
+namespace Infrastructure.Services.Development;
+
+public class NullEmailService : IEmailService
 {
-    public class NullEmailService : IEmailService
+    private readonly ILogger<NullEmailService> _logger;
+    private readonly IEmailService _emailService;
+    private readonly DevSettings _devSettings;
+
+    public NullEmailService(ILogger<NullEmailService> logger, IEmailService emailService, DevSettings devSettings)
```
con-nihr, line 14: This isn't a `NullEmailService` anymore, it actually does something. Give it a more appropriate name.

src/Infrastructure/Services/Development/NullEmailService.cs
```diff
@@ -1,16 +1,36 @@
...
+    {
+        _logger = logger;
+        _emailService = emailService;
+        _devSettings = devSettings;
+    }
+
+    public Task SendEmailAsync(string to, string subject, string body)
     {
-        public Task SendEmailAsync(string to, string subject, string body)
+        if (_devSettings.ShouldBypassEmail)
```
con-nihr, line 23: The default should be off in non-production and you have to turn it on, rather than the other way around.

src/StudyApi/Controllers/V1/DeleteController.cs
con-nihr, line 1: I would prefer that this is not actually loaded at all if it can't be accessed in non-development scenarios.

con-nihr (changes requested):

PR #62, opened 2024-04-11
Structure cleanup, particularly around common patterns in NIHR.*.

con-nihr:
- Removed unused packages.
- Consolidated package versions across projects.
- Breakup NIHR.Infrastructure to allow more flexible use.
    - You won't want to pull in MySQL dependencies if the target project doesn't uses SQL Server.
    - You won't want EntityFramework dependencies if the target project doesn't have a DB.
- Fixed a few `async` and `nullable` warnings.


BPOR/BPOR.Rms/Controllers/EmailController.cs
```diff
@@ -9,29 +9,28 @@ public Task<IActionResult> SetupCampaign()
     {
         return Task.FromResult<IActionResult>(View());
     }
-    
+
     [HttpPost]
-    public async Task<IActionResult> SetupCampaign(SetupCampaignViewModel model)
+    public IActionResult SetupCampaign(SetupCampaignViewModel model)
```
paddyduff31, line 14: Although this isn't async atm it will become

con-nihr: This is a minor one. I wanted to resolve the current build warning. This can be made async again when needed.

PR #73, opened 2024-05-16
Gov notify poc

paddyduff31:


BPOR/BPOR.Rms/Controllers/NotifyCallbackController.cs
```diff
@@ -0,0 +1,104 @@
...
+        await context.SaveChangesAsync(cancellationToken);
+
+        return Ok();
+    }
+
+    [HttpGet]
+    [Route("registerinterest")]
+    public async Task<IActionResult> RegisterInterest([FromQuery] string reference, CancellationToken cancellationToken)
+    {
+        var participant = await context.EmailCampaignParticipants
```
con-nihr, line 74: We don't want to expose the reference directly here.
An encrypted string would be preferable, so it is hidden, but we can still look it up directly in the database after decryption.

BPOR/BPOR.Rms/Controllers/NotifyCallbackController.cs
```diff
@@ -0,0 +1,104 @@
...
+                .Where(x => x.Reference == reference)
+                .Select(x => x.ParticipantId)
+                .FirstOrDefault()).FirstOrDefaultAsync(cancellationToken);
+
+        if (participant == null)
+        {
+            return NotFound();
+        }
+
+        participant.RegisteredInterestAt = DateTime.UtcNow;
```
con-nihr, line 85: We need to track down any use of fixed DateTime values and replace with a TimeProvider https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0

BPOR/BPOR.Rms/Models/Email/SetupCampaignViewModel.cs
```diff
@@ -20,7 +20,7 @@ public class SetupCampaignViewModel
 
     [DisplayName("Select email template")]
     public string? SelectedTemplateId { get; set; }
-    public IEnumerable<SelectListItem> EmailTemplates { get; set; } = [];
+    public TemplateList EmailTemplates { get; set; } = new ();
```
con-nihr, line 23: Need to abstract this away from the Notify API. It should not be visible at this level.
We are dealing with the notification API through RMS not GOVUK Notify.

BPOR/BPOR.Rms/Models/Study/StudyDetailsViewModel.cs
```diff
@@ -29,7 +29,7 @@ public class EmailCampaign
 
 public class EmailCampaignParticipant
 {
-    public string ContactEmail { get; set; }
+    public string? ContactEmail { get; set; }
```
con-nihr, line 32: Why nullable? They have to have an email in order for us to contact them.

BPOR/BPOR.Rms/Services/EmailCampaignService.cs
```diff
@@ -1,140 +1,284 @@
...
+        return refDataService.GetEmailDeliveryStatusId(EmailDeliveryStatus.Pending) ??
+               throw new InvalidOperationException("Email delivery status not found");
+    }
+
+    private async Task<FilterCriteria?> GetFilterCriteriaAsync(EmailCampaign campaign,
+        CancellationToken cancellationToken)
+    {
+        return await context.FilterCriterias
+            .AsNoTracking()
+            .IncludeAllFilterProperties()
```
con-nihr, line 65: Project this into the correct minimal data structure.

BPOR/BPOR.Rms/Services/EmailCampaignService.cs
```diff
@@ -1,140 +1,284 @@
...
+            .AsEmailCampaignParticipant()
+            .ToListAsync(cancellationToken);
+    }
+
+    private async Task ProcessAndQueueVolunteersAsync(List<EmailParticipantDetails> volunteers, EmailCampaign campaign,
+        FilterCriteria dbFilter, int emailDeliveryStatusId, CancellationToken cancellationToken)
+    {
+        const int batchSize = 1000;
+
+        var startTime = DateTime.UtcNow;
```
con-nihr, line 88: Use TimeProvider.

BPOR/BPOR.Rms/Services/EmailCampaignService.cs
```diff
@@ -1,140 +1,284 @@
...
+
+        await foreach (var processingResult in ProcessVolunteersAsync(volunteers, campaign, dbFilter,
+                           emailDeliveryStatusId, batchSize, cancellationToken))
+        {
+            await AddParticipantsToContextAsync(processingResult, cancellationToken);
+            emailQueue.Add(processingResult);
+        }
 
-            await context.SaveChangesAsync(cancellationToken);
+        var endTime = DateTime.UtcNow;
```
con-nihr, line 100: TimeProvider

BPOR/BPOR.Rms/Services/EmailCampaignService.cs
```diff
@@ -1,140 +1,284 @@
...
+        await QueueEmailSendingTaskAsync(emailQueue, campaign, cancellationToken);
+    }
+
+    private async Task SaveChangesWithRetryAsync(List<ProcessingResults> emailQueue,
+        CancellationToken cancellationToken)
+    {
+        var retryCount = 0;
+        const int maxRetries = 3;
+
+        while (retryCount < maxRetries)
```
con-nihr, line 114: Can use Polly or similar here.

BPOR/BPOR.Rms/Services/EmailCampaignService.cs
```diff
@@ -1,140 +1,284 @@
...
+        processingResult.EmailAddresses.Add(volunteer.Email);
 
-        string volunteerReference = randomNumber;
+        var personalisation = new Dictionary<string, dynamic?>
+        {
+            { "email", volunteer.Email },
+            { "emailCampaignParticipantId", emailCampaignParticipant.Id },
+            { "firstName", volunteer.FirstName },
+            { "lastName", volunteer.LastName },
+            { "uniqueLink", $"{appSettings.Value.BaseUrl}/NotifyCallback/registerinterest?reference={reference}" },
```
con-nihr, line 205: Use helper methods to generate accurate URL to action.

con-nihr: Don't expose the reference here directly.

BPOR/BPOR.Rms/Settings/AppSettings.cs
```diff
@@ -0,0 +1,17 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace BPOR.Rms.Settings;
+
+public class AppSettings : IValidatableObject
+{
+    public static string SectionName => "AppSettings";
+    [Required] public string BaseUrl { get; set; } = string.Empty;
```
con-nihr, line 8: We can probably get this at runtime from the web host.

BPOR/BPOR.Rms/Startup/DependencyInjection.cs
```diff
@@ -52,31 +53,45 @@ public static class DependencyInjection
                 builder.UseQuerySplittingBehavior(QuerySplittingBehavior.SplitQuery);
             }));
 
-        var logger = services.BuildServiceProvider().GetService<ILoggerFactory>()
-            ?.CreateLogger("BPOR.Rms");
+        var logger = services.BuildServiceProvider().GetService<ILoggerFactory>()?.CreateLogger("BPOR.Rms");
```
con-nihr, line 56: Need to look at this. We shouldn't have to build the service provider so early.

BPOR/BPOR.Rms/Utilities/ReferenceGenerator.cs
```diff
@@ -0,0 +1,24 @@
...
+    private static readonly Random _random = new();
+    
+    public string GenerateReference()
+    {
+        var reference = new StringBuilder(16);
+        Span<int> span = stackalloc int[15];
+        for (var i = 0; i < span.Length; i++)
+        {
+            span[i] = _random.Next(0, 10);
+            reference.Append(span[i]);
```
con-nihr, line 18: The span is good here, but it is not really being used. This could be optimised.

BPOR/BPOR.Rms/Views/Email/Partials/_TemplateSelect.cshtml
```diff
@@ -16,7 +16,7 @@
     }
 </style>
 
-<div class="govuk-form-group @(Html.ViewData.ModelState["SelectedTemplateId"]?.Errors.Count > 0 ? "govuk-form-group--error" : "")" asp-for="SelectedTemplateId">
+<div class="govuk-form-group" @(Html.ViewData.ModelState["SelectedTemplateId"]?.Errors.Count > 0 ? "govuk-form-group--error" : "") asp-for="SelectedTemplateId">
```
con-nihr, line 19: This doesn't look correct. `govuk-form-group--error` should be added as a class.

BPOR/NIHR.NotificationService/Models/SendBatchEmailRequest.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Concurrent;
+
+namespace NIHR.NotificationService.Models
+{
+    public class SendBatchEmailRequest
+    {
+        public IEnumerable<string> EmailAddresses { get; set; }
```
con-nihr, line 7: I would pair the email with the personalisation data so they don't get mixed up during processing.

BPOR/NIHR.NotificationService/Models/SendEmailRequest.cs
```diff
@@ -0,0 +1,9 @@
+namespace NIHR.NotificationService.Models;
+
+public class SendEmailRequest
+{
+    public string EmailAddress { get; set; }
```
con-nihr, line 5: Same, email address is probably redundant here.

con-nihr (approved):

BPOR/BPOR.Geolocation/Controllers/GeolocationController.cs
```diff
@@ -1,54 +1,88 @@
...
     {
-        var participants = await context.Participants
-            .Include(p => p.Address)
-            .Take(20)
-            .ToListAsync(cancellationToken);
+        var cache = new ConcurrentDictionary<string, (double Latitude, double Longitude)>();
+        var semaphore = new SemaphoreSlim(600, 600);
+        var tasks = new List<Task>();
+
+        await foreach (var participant in GetParticipantsWithAddressesAsync(cancellationToken))
```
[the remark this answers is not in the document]
paddyduff31: Have had to pull into in memory list to do this, not ideal but we can discuss

BPOR/BPOR.Rms/Controllers/NotifyCallbackController.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return Unauthorized();
+        }
+
+        if (message.Reference == "PreviewEmailReference")
+        {
+            return Ok();
+        }
+
+        if (!int.TryParse(message.Reference, out var participantEmailId))
```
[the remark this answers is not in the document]
paddyduff31: so this is the id from the email participant id and not the generated reference but all other comments around this are valid

BPOR/BPOR.Rms/Utilities/ReferenceGenerator.cs
paddyduff31: yeah, I was trying to take your feedback on creating an array of numbers but then converting to a string but I found myself looping twice which was a waste and the luhn CalculateCheckDigit expects a string array so was all a bit of a mess

Have reverted to string builder only and we can revisit

BPOR/NIHR.NotificationService/Models/SendBatchEmailRequest.cs
paddyduff31: so currently this is the dictionary key, should we leave it there and just use it?  Doesn't seem very explicit though and if we want to use the notification service and not pass any personalisation data?

BPOR/BPOR.Rms/Controllers/NotifyCallbackController.cs
paddyduff31: Will look at in it's own PR as there will be a lot changed

PR #74, opened 2024-05-30
Use rbec postcode validation

paul-tennyson:


BPOR/BPOR.Rms/Models/Filter/Postcode.cs
con-nihr, line 1: Pull this in as a nuget package
`dotnet add package Rbec.Postcodes --version 1.0.0`

con-nihr (changes requested):

BPOR/BPOR.Rms/Models/Filter/Postcode.cs
paul-tennyson: @con-nihr That's updated now with the NuGet package

con-nihr (approved):

PR #86, opened 2024-06-28
moved email service to shared

paddyduff31:


paddyduff31:
@con-nihr still a bit of work to tidy up but this can circled back to as I think moving any contentful work out of common and into the new NIHR infra is probably the way to go but you can use `IEmailService` to send an SES email and get the full email content(including styles and template etc) using `IContetfulService`

PR #97, opened 2024-07-30
Add tests for error view context extension

petecorriepa:
- Added test project (As project specific to support moving to shared package in the future)
- Added some tests for order errors

### DIFF
--- a/BPOR/BPOR.sln
+++ b/BPOR/BPOR.sln
@@ -35,6 +35,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ImportGeocodeData", "Import
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NIHR.GovUk.AspNetCore.Mvc", "NIHR.GovUk.AspNetCore.Mvc\NIHR.GovUk.AspNetCore.Mvc.csproj", "{310FC69F-836B-4940-9634-EE74BA4F219C}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NIHR.GovUk.AspNetCore.Mvc.Tests", "NIHR.GovUk.AspNetCore.Mvc.Tests\NIHR.GovUk.AspNetCore.Mvc.Tests.csproj", "{076E25AB-51AA-4622-9522-874D59275E96}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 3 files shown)


con-nihr (approved):
Love it.

Not necessarily for this PR but I would like to extend this with more than two levels of nesting and tag helper rendering tests for Error Summary and Form Group to make sure the errors are presented correctly. This can be done in unit tests so we don't have to resort to e2e testing.

petecorriepa:
> Love it.
> 
> Not necessarily for this PR but I would like to extend this with more than two levels of nesting and tag helper rendering tests for Error Summary and Form Group to make sure the errors are presented correctly. This can be done in unit tests so we don't have to resort to e2e testing.

Added one more test to include one more level of testing - will have a play with the rendering tests, I did try but got blocked by a nested dependency - will look into it!

con-nihr:
> > Love it.
> > Not necessarily for this PR but I would like to extend this with more than two levels of nesting and tag helper rendering tests for Error Summary and Form Group to make sure the errors are presented correctly. This can be done in unit tests so we don't have to resort to e2e testing.
> 
> Added one more test to include one more level of nesting - will have a play with the rendering tests, I did try but got blocked by a nested dependency - will look into it!

Do the error messages appear in a different order 'natively' and are the tests establishing that the correct result is because of the code under test rather than just a co-incidence based on the name of the fields?

petecorriepa:
> > > Love it.
> > > Not necessarily for this PR but I would like to extend this with more than two levels of nesting and tag helper rendering tests for Error Summary and Form Group to make sure the errors are presented correctly. This can be done in unit tests so we don't have to resort to e2e testing.
> > 
> > 
> > Added one more test to include one more level of nesting - will have a play with the rendering tests, I did try but got blocked by a nested dependency - will look into it!
> 
> Do the error messages appear in a different order 'natively' and are the tests establishing that the correct result is because of the code under test rather than just a co-incidence based on the name of the fields?

The tests are trying to say its irrelevant what happens natively - we want it to be always driven off display order attribute
E.g. 'Error message xx' I was trying to illustrate that just because one happened 'first' it doesn't mean we display it that way - i.e. the key and the display order attribute drive the order - not the error message itself - is that what you mean?

con-nihr:
> > > > Love it.
> > > > Not necessarily for this PR but I would like to extend this with more than two levels of nesting and tag helper rendering tests for Error Summary and Form Group to make sure the errors are presented correctly. This can be done in unit tests so we don't have to resort to e2e testing.
> > > 
> > > 
> > > Added one more test to include one more level of nesting - will have a play with the rendering tests, I did try but got blocked by a nested dependency - will look into it!
> > 
> > 
> > Do the error messages appear in a different order 'natively' and are the tests establishing that the correct result is because of the code under test rather than just a co-incidence based on the name of the fields?
> 
> The tests are trying to say its irrelevant what happens natively - we want it to be always driven off display order attribute E.g. 'Error message xx' I was trying to illustrate that just because one happened 'first' it doesn't mean we display it that way - i.e. the key and the display order attribute drive the order - not the error message itself - is that what you mean?

I'm thinking the display order is very close to what the native order would be anyway.
Ideally the display order would be radically different that what would happen with the normal error messages so we can be sure that we're seeing the effect of this code.

con-nihr:
@petecorriepa FYI
This is how the framework sets up TagHelper tests:
https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.TagHelpers/test/LabelTagHelperTest.cs#L152

Nothing to do for this PR but we will need to get especially good coverage for the components we are making widely reusable.

PR #107, opened 2024-08-08
CRNCC-2275

Tkirkwood:
added method to return a string in the common uk format for display.

### DIFF
--- a/BPOR/BPOR.Rms/Models/Researcher/ResearcherStudyFormViewModel.cs
+++ b/BPOR/BPOR.Rms/Models/Researcher/ResearcherStudyFormViewModel.cs
@@ -83,8 +83,19 @@ namespace BPOR.Rms.Models.Researcher
                 return null;
             }
         }
+        private static String? formatDateForUK(int? year, int? month, int? day)
+        {
+            if (!year.HasValue || !month.HasValue || !day.HasValue)
+            {
+                return null;
+            }
+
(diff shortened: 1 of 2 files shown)


BPOR/BPOR.Rms/Models/Researcher/ResearcherStudyFormViewModel.cs
```diff
@@ -83,9 +83,20 @@ public class GovUkDate
...
+            }
+
+            return ""+ day.Value + "/" + month.Value + "/" + year.Value;
+        }
 
         public DateOnly? ToDateOnly() => ConstructDate(Year, Month, Day);
 
+        public String? UKDisplayDate() => formatDateForUK(Year, Month, Day);
+
         public static GovUkDate FromDateTime(DateTime? date) => new GovUkDate { Day = date?.Day, Month = date?.Month, Year = date?.Year };
```
paul-tennyson, line 100: @Tkirkwood Are we not able to just use this GovUkDate to display the dates in the UK format? Or does this still display in US format on the frontend?

Tkirkwood: using the toDateOnly displays in theUS mm/dd/yyyy format.

paul-tennyson (approved):

PR #114, opened 2024-08-26
CRNCC-2286 cancel away from editing dependent

Tkirkwood:
model Id was not passed in method so the dependant question did not have it to link back to in both the cancel button as well as the breadcrumbs.

https://nihr.atlassian.net/browse/CRNCC-2286

### DIFF
--- a/BPOR/BPOR.Rms/Controllers/ResearcherController.cs
+++ b/BPOR/BPOR.Rms/Controllers/ResearcherController.cs
@@ -472,6 +472,7 @@ public class ResearcherController(ParticipantDbContext context, ICurrentUserProv
         "HasFunding,FundingCode,UKRecruitmentTarget,TargetPopulation,RecruitmentStartDate,RecruitmentEndDate,RecruitingIdentifiableVolunteers,Step")]
         ResearcherStudyFormViewModel model)
     {
+        model.Id = id;
         model.PortfolioSubmissionStatusOptions = context.Submitted.ToList();
         model.OutcomeOfSubmissionOptions = context.SubmissionOutcome.ToList();
 
@@ -479,7 +480,6 @@ public class ResearcherController(ParticipantDbContext context, ICurrentUserProv
 
         ValidateMandatoryFields(model);
 
-        model.Id = id;
         switch (model.Step)
         {
             case 2:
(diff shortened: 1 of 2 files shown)


paul-tennyson (approved):

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
```diff
@@ -468,7 +468,7 @@ public async Task<IActionResult> Edit(int id, int field)
     [HttpPost]
     [ValidateAntiForgeryToken]
     public async Task<IActionResult> Edit(int id,
-        [Bind("ShortName,ChiefInvestigator,StudySponsors,CPMSId,PortfolioSubmissionStatus,OutcomeOfSubmission," +
+        [Bind("ShortName,ChiefInvestigator,StudySponsors,CPMSId,PortfolioSubmissionStatus,OutcomeOfSubmission,Id," +
```
con-nihr, line 471: If we need to `Bind` it is probably a sign the model is too complicated.
The model should only include properties that we want to bind anyway.

`Id` is already passed in as a route parameter so arguably we should have a few different models depending on what is needed.

con-nihr (changes requested):
I'll support some of these changes to restructure the model as needed.

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
Tkirkwood: I went back over the method and noticed I overlooked a line near the end of the code
`model.Id = id;` 
this is not reached when going through a dependancy question as that returns much further up.
would it be better to remove the bound id and move this line nearer the start of the method, or is a restructure for the model still the better option?

con-nihr: I think we could benefit from a slight larger restructuring of things in this area, but if we want to keep things compact to close out this fix I would be fine with moving `model.Id = id;` right up to the first line of the method.

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
```diff
@@ -479,6 +479,7 @@ public async Task<IActionResult> Edit(int id,
 
         ValidateMandatoryFields(model);
 
+        model.Id = id;
```
con-nihr, line 482: I would literally put this as the first line.

There may be future changes to `ValidateMandatoryFields()` that might expect `model.Id` to be populated.

con-nihr (changes requested):

con-nihr (approved):

PR #115, opened 2024-08-27
Rms fix/crncc 2293-page-titles

Tkirkwood:
https://nihr.atlassian.net/browse/CRNCC-2293

updated page titles,
created method in researcher controller for page titles in the apply to add study route.

### DIFF
--- a/BPOR/BPOR.Rms/Controllers/ResearcherController.cs
+++ b/BPOR/BPOR.Rms/Controllers/ResearcherController.cs
@@ -34,6 +34,7 @@ public class ResearcherController(ParticipantDbContext context, ICurrentUserProv
 
     public IActionResult SubmitStudy(ResearcherStudyFormViewModel model)
     {
+        ViewData["Title"] = getPageTitle(model.Step);
         ViewData["Step"] = model.Step;
         ViewData["TotalSteps"] = model.TotalSteps;
         return View(model);
@@ -170,10 +171,40 @@ public class ResearcherController(ParticipantDbContext context, ICurrentUserProv
         model.PortfolioSubmissionStatusOptions = context.Submitted.ToList();
         model.OutcomeOfSubmissionOptions = context.SubmissionOutcome.ToList();
 
+
+        ViewData["Title"] = getPageTitle(model.Step);
         ViewData["Step"] = model.Step;
         ViewData["TotalSteps"] = model.TotalSteps;
         return View(model);
     }
+
+    private String getPageTitle(int? step)
+    {
+        if (step == null)
+        {
+            return "Submit study";
+        }
+        switch (step)
+        {
+            case 1:
+                return "Study information";
+            case 2:
+            case 3:
+                return "NIHR CRN portfolio status ";
+            case 4:
+            case 5:
+                return "NIHR funding";
+            case 6:
+                return "Target population";
+            case 7:
+                return "Study recruitment";
+            case 8:
+                return "Check your answers";
+            default:
+                return "Submit study";
+        }
+    }
+
 
     private void ValidateMandatoryFields(ResearcherStudyFormViewModel model)
     {
--- a/BPOR/BPOR.Rms/Controllers/ResearcherController.cs
+++ b/BPOR/BPOR.Rms/Controllers/ResearcherController.cs
@@ -34,7 +34,6 @@ public class ResearcherController(ParticipantDbContext context, ICurrentUserProv
 
     public IActionResult SubmitStudy(ResearcherStudyFormViewModel model)
     {
-        ViewData["Title"] = getPageTitle(model.Step);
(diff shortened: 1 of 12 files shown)


paul-tennyson (approved):

petecorriepa (approved):

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
```diff
@@ -170,11 +171,41 @@ public async Task<IActionResult> SubmitStudy(ResearcherStudyFormViewModel model,
...
         model.OutcomeOfSubmissionOptions = context.SubmissionOutcome.ToList();
 
+
+        ViewData["Title"] = getPageTitle(model.Step);
         ViewData["Step"] = model.Step;
         ViewData["TotalSteps"] = model.TotalSteps;
         return View(model);
     }
 
+    private String getPageTitle(int? step)
```
con-nihr, line 181: It seems like having a different page title for each step could be a common pattern for multi-step forms.

Extend the abstract class `FormWithSteps` with an abstract property for the step name:
`public abstract string StepName { get; }`

Implement that property for the specific below case in `ResearcherStudyFormViewModel`.

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
```diff
@@ -34,6 +34,7 @@ public IActionResult TermsAndConditions(ResearcherTermsAndConditionsViewModel mo
 
     public IActionResult SubmitStudy(ResearcherStudyFormViewModel model)
     {
+        ViewData["Title"] = getPageTitle(model.Step);
```
con-nihr, line 37: I don't think we should be manipulating `ViewData` here to initialise a multi-step form.
We can do all of this in the view by accessing the model directly, no need to do this twice in the two `SubmitStudy` actions.

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
```diff
@@ -170,11 +171,41 @@ public async Task<IActionResult> SubmitStudy(ResearcherStudyFormViewModel model,
...
         ViewData["Step"] = model.Step;
         ViewData["TotalSteps"] = model.TotalSteps;
         return View(model);
     }
 
+    private String getPageTitle(int? step)
+    {
+        if (step == null)
+        {
+            return "Submit study";
```
con-nihr, line 185: you could just rely on the `default` case here.

BPOR/BPOR.Rms/Views/Shared/_Layout.cshtml
```diff
@@ -6,7 +6,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta name="x-environment" content="@HostEnvironment.EnvironmentName" />
-    <title>@ViewData["Title"] - Be Part of Research Volunteer Registry</title>
+    <title>@ViewData["Title"] - Be Part of Research volunteer registry - Be Part of Research</title>
```
con-nihr, line 9: Is this what the client has requested?

It seems redundant to duplicate `Be Part of Research` in the title.

What is the reasoning behind this title format?

con-nihr (changes requested):
Do we also need to update the page title in `BPOR\BPOR.Rms\Views\Researcher\Edit.cshtml` for the `ResearcherStudyFormViewModel` steps?

Can we align the page titles and step names so you can use the new `StepName` property to set the title and this heading in each step?

https://github.com/PA-NIHR-CRN/dte-study-api/blob/ccb0baded8992f765a8bdd1dbccc10b1d7d5afef/BPOR/BPOR.Rms/Views/Researcher/Partials/_Step1FormFields.cshtml#L3-L5

BPOR/BPOR.Rms/Views/Shared/_Layout.cshtml
Tkirkwood: this is the format that was requested, I am not aware of the specific reasoning, but I believe they are following the format of this example https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G88#G88-ex1 from wcag 2.0

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
Tkirkwood: removed the null check, now just relies on the default of the switch statement.

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
Tkirkwood: implemented property for `FormWithSteps` this also had a knock on effect for the researcher pages ( a 2 step form ).  changed the edit pages on client feedback to use the same stepNames as well.

PR #118, opened 2024-09-02
Fix rms/crncc 2256-progress-indicators

Tkirkwood:
removed progress indicator, 'question x of y'
changed progress bar to only show when there are 3 or more pages.


kevinofficer (approved):

con-nihr (commented):
The original requirements don't seem to be GDS compliant.

Hold off until we have clarification.

We might just remove the progress bar and question indicator text completely.

con-nihr (changes requested):
On hold pending clarification.

BPOR/BPOR.Rms/Views/Researcher/SubmitStudy.cshtml
```diff
@@ -10,15 +10,7 @@
     <div class="govuk-grid-column-two-thirds govuk-grid-column-one-half-from-desktop">
 
         <form asp-action="SubmitStudy" back-link-for="Step">
-            @if (ViewData["Step"] is not null && ViewData["TotalSteps"] is not null)
-            {
-                var step = ViewData["Step"] as int?;
```
con-nihr, line 15: Check where else we are using `ViewData["Step"]` and `ViewData["TotalSteps"]` if it is not read _anywhere_ anymore, remove all the locations that are setting this these values.

e.g.
https://github.com/PA-NIHR-CRN/dte-study-api/pull/115/files#diff-a27974a40fdd743b11a3a7b0675ac1d489fdd14c19f875f6c8487e7c0362c0c6R174

con-nihr (approved):

PR #120, opened 2024-09-18
Prototype volunteer registration

paul-tennyson:


BPOR/BPOR.Rms/Controllers/VolunteerController.cs
```diff
@@ -11,6 +12,123 @@ namespace BPOR.Rms.Controllers;
...
+
+        if (ModelState.IsValid)
+        {
+            return RedirectToAction(nameof(VolunteerController.AccountSuccess));
+        }
+
+        return View(model);
+    }
+
+    private void DoesPostcodeSurnameDoBComboExist(string postCode, string lastName, GovUkDate dateOfBirth)
```
petecorriepa, line 82: this method can be marked as async and return a Task
then use await (line 86).... FirstOrDefaultAsync() (line 90)

BPOR/BPOR.Rms/Controllers/VolunteerController.cs
```diff
@@ -11,6 +12,123 @@ namespace BPOR.Rms.Controllers;
...
+        p.Address.Postcode == postCode)
+        .FirstOrDefault();
+
+        if (user != null)
+        {
+            ModelState.AddModelError("LastName", "Combination of surname, date of birth and postcode already exists and cannot be used");
+        }
+    }
+
+    private void DoesUserEmailExistInDatabase(string email)
```
petecorriepa, line 98: same here, this can be async

BPOR/BPOR.Rms/Controllers/VolunteerController.cs
```diff
@@ -11,6 +12,123 @@ namespace BPOR.Rms.Controllers;
...
+    private void ValidateDateOfBirth(GovUkDate dateOfBirth)
+    {
+        if (!dateOfBirth.HasValue)
+        {
+            ModelState.AddModelError("DateOfBirth", "Date of birth is required");
+        }
+
+        if (dateOfBirth.HasValue)
+        {
+            DateOnly today = DateOnly.FromDateTime(DateTime.Today);
```
petecorriepa, line 117: should this be >=?
I can sign up on my 18th birthday

BPOR/BPOR.Rms/Controllers/VolunteerController.cs
paul-tennyson: Thanks @petecorriepa I've updated these now to use async tasks

BPOR/BPOR.Rms/Controllers/VolunteerController.cs
paul-tennyson: The current method actually does accept a date that would on on your 18th birthday. If I input 18 09 2006, it accepts this. If I input 19 09 2006, it flags as a fail.

petecorriepa: ok my bad!

petecorriepa (approved):

PR #138, opened 2025-02-05
Aria label updates for campaign page

paul-tennyson:


Tkirkwood (approved):

con-nihr (commented):
I would challenge the need for `aria-label` at all when it is redundant because its value matches the inner text for the link.

We only need an `aria-label` if it adds additional information to the link text.

JoeDonaghy-PA (approved):

paul-tennyson:
@con-nihr I've updated this to remove unnecessary aria labels, and updated others with more context to use the welsh translation

PR #135, opened 2024-11-26
Letter campaign combined tickets.

Tkirkwood:
https://nihr.atlassian.net/browse/CRNCC-2387
https://nihr.atlassian.net/browse/CRNCC-2393
https://nihr.atlassian.net/browse/CRNCC-2394
https://nihr.atlassian.net/browse/CRNCC-2396
https://nihr.atlassian.net/browse/CRNCC-2416

PR #140, opened 2025-03-20
bug: update ethnic other, add optional label, update page title,  cle…

kevinofficer:
…ar postcode

### DIFF
--- a/BPOR/BPOR.Rms/Constants/EthnicBackgrounds.cs
+++ b/BPOR/BPOR.Rms/Constants/EthnicBackgrounds.cs
@@ -6,7 +6,6 @@ namespace BPOR.Rms.Constants
         public EthnicBackgrounds()
         {
         }
-        public readonly static String CommonOther = "Other";
 
         private static List<Dictionary<string, string>> Asian = new List<Dictionary<string, string>>
         {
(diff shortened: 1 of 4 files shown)


BPOR/BPOR.Rms/Constants/EthnicBackgrounds.cs
```diff
@@ -23,28 +22,28 @@ public EthnicBackgrounds()
...
             new Dictionary<string, string> { { "label", "Irish Traveller" }, { "value", "Irish Traveller" } },
             new Dictionary<string, string> { { "label", "Roma" }, { "value", "Roma" } },
-            new Dictionary<string, string> { { "label", "Another White background" }, { "value", CommonOther } }
+            new Dictionary<string, string> { { "label", "Another White background" }, { "value", "Another White background" } }
         };
         private static List<Dictionary<string, string>> Other = new List<Dictionary<string, string>>
         {
             new Dictionary<string, string> { { "label", "Arab" }, { "value", "Arab" } },
-            new Dictionary<string, string> { { "label", "Any other ethnic group" }, { "value", CommonOther } }
+            new Dictionary<string, string> { { "label", "Any other ethnic group" }, { "value", "Any other ethnic group" } }
```
kevinofficer, line 46: Chris has a comment elsewhere to change this. the changes is just an immediate fix.

JoeDonaghy-PA: Just checking, would this lead to exiting values in the database no longer mapping? Is there an update script for these values, or would it have any impact on the UI?

kevinofficer: the db values are saved as the longer version rather than `CommonOther`. This is impacting how they display on the UI

Here is the recommendations from Chris too, but I'll be addressing them in a later PR
https://github.com/PA-NIHR-CRN/dte-study-api/pull/135/files#r1967601491

JoeDonaghy-PA (approved):

con-nihr (approved):

PR #156, opened 2025-05-01
feat: updated successful letters to delivered.

kevinofficer:


Tkirkwood (approved):

kira-clarke (commented):
My immediate concern is that UpdateDeliveryStatusForLetterAsync can throw an exception in response to an invalid value, and doing so will cause the db change saves in HostedNotificationQueueService.cs L75/76 to be skipped and so the IsProcessed values are never updated, so on the next iteration 10 seconds later the same records will be processed again, and very possibly fail in the same manner.

The exception that is thrown on HostedNotificationQueueService.cs L103 does not identify the notification that failed so the log entry written on L81 will not be as helpful as it could be in diagnosing the fault.

From a coding style pov - the only thing I'd change would be using constant strings rather than literals to improve readability/maintainability.

This just from a first read through - I've not looked at any unmodified code to get more context yet!

BPOR/NIHR.NotificationService/Services/HostedNotificationQueueService.cs
```diff
@@ -80,5 +86,37 @@ public override async Task StopAsync(CancellationToken stoppingToken)
...
+                    campaignParticipantIdValue);
+                return;
+            }
+
+            var participant = await context.CampaignParticipant.FirstOrDefaultAsync(x => x.Id == campaignParticipantId);
+            if (participant == null)
+            {
+                _logger.LogWarning("Notification ID {NotificationId}: Participant ID {ParticipantId} not found", notification.Id, campaignParticipantId);
+                return;
+            }
```
kevinofficer, line 116: @kira-clarke addressed your first bit of feedback regarding exceptions. I think it's wrong to throw exceptions here anyway, we could have thousands of notifications to process and we dont want an exception to prevent processing of the remaining queue.
let me know what you think

BPOR/BPOR.Domain/Constants/PersonalisationKeys.cs
```diff
@@ -0,0 +1,14 @@
...
+public static class PersonalisationKeys
+{
+    public const string CampaignParticipantId = "campaignParticipantId";
+    public const string CampaignTypeId = "campaignTypeId";
+    public const string FirstName = "firstName";
+    public const string LastName = "lastName";
+    public const string UniqueLink = "uniqueLink";
+    public const string TemplateId = "templateId";
+    public const string UniqueReference = "uniqueReference";
+    public const string Email = "email";
```
kevinofficer, line 12: @kira-clarke done some refactoring for all personalisation keys

BPOR/NIHR.NotificationService/Services/HostedNotificationQueueService.cs
```diff
@@ -93,23 +92,32 @@ private async Task UpdateDeliveryStatusForLetterAsync(Notification notification,
...
+                    campaignParticipantIdValue);
+                return;
+            }
+
+            var participant = await context.CampaignParticipant.FirstOrDefaultAsync(x => x.Id == campaignParticipantId);
+            if (participant == null)
+            {
+                _logger.LogWarning("Notification ID {NotificationId}: Participant ID {ParticipantId} not found", notification.Id, campaignParticipantId);
+                return;
             }
```
kira-clarke, line 117: That looks better! I agree with removing the exception ... I'm assuming that we monitor the logs and these warnings are acted upon?

I didn't spot the possible need for a second log entry - good catch :-)

kevinofficer: we don't proactively monitor this, but its something we can proactively monitor for the first campaigns sent after the release, then just logs we can review in the event of an incident

kira-clarke (approved):

PR #157, opened 2025-05-02
feat: resolve heading levels

kevinofficer:
previously removed labels in legends causing some css to be lost


BPOR/BPOR.Rms/Views/Shared/Partials/_AreasOfResearch.cshtml
```diff
@@ -1,30 +1,40 @@
...
+        shouldShowHintText = false;
+    }
+ }
 
 <div class="govuk-form-group">
     <fieldset class="govuk-fieldset">
         <@headingLevel  class="@headingClass" >
             <legend>
-                @displayName
+                @if (labelLevel != null)
```
JoeDonaghy-PA, line 28: This could create a heading within a heading which isn't allowed and can cause issues with screen readers. You could get around that by adding an additional check on this that the @headinglevel isn't an H tag as well

JoeDonaghy-PA (changes requested):

JoeDonaghy-PA (approved):

PR #171, opened 2025-05-15
separate sql files for migrations

kevinofficer:


.github/workflows/bpor_rms_build.yml
```diff
@@ -84,12 +84,19 @@ jobs:
         run: |
           aws ec2 authorize-security-group-ingress --group-id ${{ secrets.RDS_SG }} --ip-permissions '[{"IpProtocol": "tcp", "FromPort": 3306, "ToPort": 3306, "IpRanges": [{"CidrIp": "${{ steps.publicip.outputs.ip }}/32", "Description": "GitHub runner IP for Sponsor Engagement automation DB migration"}]}]'
  
+      - name: Confirm SQL migration files case sensitivity
+        run: |
+          if find BPOR/BPOR.Domain/Migrations/Scripts -name '*.Up.sql' -o -name '*.Down.sql' | grep -q .; then
```
con-nihr, line 89: I didn't release this was a problem we were facing. Though of course it would be on case-sensitive file systems.

I think the best way to fix this is to make the extension method support any case.

BPOR/BPOR.Domain/Migrations/Extensions/MigrationBuilderExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+            string basePath = Path.Combine(AppContext.BaseDirectory, "Migrations", "Scripts", scriptIdentifier);
+            string directionSuffix = direction.ToString().ToLowerInvariant();
+
+
+            var pattern = !string.IsNullOrEmpty(stepIdentifier)
+                ? $"{scriptIdentifier}.{stepIdentifier}.{directionSuffix}.sql"
+                : $"{scriptIdentifier}*.{directionSuffix}.sql";
+
+            var matchingFiles = Directory
+                .EnumerateFiles(basePath, "*.sql", SearchOption.TopDirectoryOnly)
```
con-nihr, line 78: This enumeration can be made case-insensitive.

https://learn.microsoft.com/en-us/dotnet/api/system.io.enumerationoptions?view=net-9.0#:~:text=or%20SecurityException).-,MatchCasing,Gets%20or%20sets%20the%20case%2Dmatching%20behavior.,-MatchType

BPOR/BPOR.Domain/BPOR.Domain.csproj
```diff
@@ -45,6 +44,9 @@
     </ItemGroup>
 
     <ItemGroup>
-      <Folder Include="Constants\" />
+      <Content Include="Migrations/Scripts/**/*.sql">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
con-nihr, line 48: I don't think this is required.

The SQL is incorporated into the migrations at build time. They are not needed at runtime.

`<CopyToOutputDirectory>Never</CopyToOutputDirectory>` should be fine.

con-nihr (changes requested):

BPOR/BPOR.Domain/BPOR.Domain.csproj
con-nihr: I see, the extension method is using `AppContext.BaseDirectory` which will be based on the output directory.

Probably best to use the current working directory instead.

.github/workflows/bpor_rms_build.yml
```diff
@@ -84,12 +84,19 @@ jobs:
...
+            exit 1
+          fi
+        
       - name: Update Database
         env:
           DB_CONNECTION_STRING: ${{ secrets.DB_CONNECTION_STRING }}
         run: |
           cd BPOR/BPOR.Domain
-          dotnet ef database update --connection "${{ env.DB_CONNECTION_STRING }}"
+          dotnet ef database update --configuration Release --connection "${{ env.DB_CONNECTION_STRING }}"
```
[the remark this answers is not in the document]
kevinofficer: ah i see what you mean, it may not be necessary to even build them to the output directory.

PR #173, opened 2025-05-16
Add Canonical Town To Reporting In Place Of Town

JoeDonaghy-PA:
Added a new field to the participant address table, 'CanonicalTown'
Update the view to use the new canonical town field 
Update the stream handler to populate the CanonicalTown field is possible on insert/modify

### DIFF
--- a/BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
+++ b/BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
@@ -6,6 +6,8 @@ using BPOR.Domain.Entities;
 using BPOR.Registration.Stream.Handler.Mappers;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
+using NIHR.Infrastructure;
+using NIHR.Infrastructure.Models;
 using static Amazon.Lambda.DynamoDBEvents.StreamsEventResponse;
 
 namespace BPOR.Registration.Stream.Handler.Handlers;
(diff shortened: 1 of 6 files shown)


BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
```diff
@@ -119,6 +122,17 @@ private async Task<Participant> InsertAsync(Dictionary<string, AttributeValue> i
         {
             // No linked participant exists, create a new one.
             targetParticipant = participantDbContext.Participants.Add(new Participant()).Entity;
+        }
+
+        if (targetParticipant.Address is not null && !String.IsNullOrWhiteSpace(targetParticipant.Address.Postcode))
+        {
+            IEnumerable<PostcodeAddressModel> addressModels;
+            addressModels = await locationApiClient.GetAddressesByPostcodeAsync(targetParticipant.Address.Postcode, new CancellationToken());
```
Tkirkwood, line 130: if we move this into the participant mapper https://github.com/PA-NIHR-CRN/dte-study-api/blob/rework-added-RMS/BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs#L157 next to where we get the add the co-ordinates we can avoid duplicate code and adding the location service.

Tkirkwood (changes requested):

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
JoeDonaghy-PA: Done 👍

PR #178, opened 2025-05-30
adding canonical town to offline journey route

Tkirkwood:
added canonical town to offline Journey.

### DIFF
--- a/BPOR/BPOR.Rms/Controllers/VolunteerController.cs
+++ b/BPOR/BPOR.Rms/Controllers/VolunteerController.cs
@@ -105,11 +105,11 @@ public class VolunteerController(ParticipantDbContext context,
                 await DoesUserEmailExistInDatabaseAsync(model.EmailAddress);
             }
 
+            string canonicalTown = null;
             if (ModelState.IsValid) { 
 
                 if (!model.ManualAddressEntry)
-                {
-
+                { 
                     if (model.SelectedAddress != null)
                     {
                         var participantAddress =  JsonSerializer.Deserialize<PostcodeAddressModel>(model.SelectedAddress);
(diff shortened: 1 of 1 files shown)


BPOR/BPOR.Rms/Controllers/VolunteerController.cs
```diff
@@ -122,9 +122,21 @@ public async Task<IActionResult> Create(VolunteerFormViewModel model, string act
                         if (Postcode.TryParse(participantAddress.Postcode, out TempPostcode))
                         {
                             model.PostCode = TempPostcode;
+                            List<PostcodeAddressModel> possibleAddresses = await GetAddresses(TempPostcode.ToString());
+                            if (possibleAddresses.Count > 0) {
+                                canonicalTown = possibleAddresses.First().Town;
+                            }
```
kevinofficer, line 128: looks like `canonicalTown` is only set on postcode lookup. its set here before the address is selected from a dropdown of address.count > 0, or when an address is selected but it doesn't cover cases where a user has entered an address manually, or not used the postcode lookup.

this currently won't solve the problem. users can provide an invalid postcode manually by either entering address manually and providing a bad town, or they could update the town freely after using the postcode service. Both result in bad town data.

Its worth discussing further but my initial thoughts are
- canonicalTown is potentially not even required, if we just use Town for reporting. if they need to be separate for some reason thats fine, but lets understand why we are recording town twice.
- maybe we shouldn't be displaying town in the UI if we want to protect it from bad user input.
- we could assign town in the UI, but hidden, or assign it at the submission stage where the user input doesn't impact town.
- for manually entered addresses, what happens if the postcode is invalid and we don't find any postcode data? the vs location api currently expects a full UK postcode, otherwise the results are 0, and the canonicalTown will remain as null.
- we could explore expanding the vs location api to allow us to search partial postcodes via OS e.g. 
```https://api.os.uk/search/places/v1/find?query=BT14&key=####```
`v1/find` is a free text search that we can use to search partial postcodes with the Town as part of the response, without relying on a full valid postcode.

Comment (kevinofficer) [reply BPOR/BPOR.Rms/Controllers/VolunteerController.cs:128]: Thanks. I've had a read of the ticket and yes, what I've suggested is much more than the ask, but I believe we still aren't meeting the requirements of the ticket, with some of my above points still valid.

The tickets suggested approach is
```
Update the stream handler process (which passes new VS registrations to the reporting DB) to use the postcode entered / captured by the volunteer to lookup (via the location service) and populate CanonicalTown with a standardised value for Town, rather than rely on the possibly manually entered value in ParticipantAddress.Town, which is where the issue lies currently.
```

This is my understanding of the ticket
- if a user uses the postcode lookup service, do nothing, we assume the Town is correct or standardised.
- if a user manually enters the address data, we want to assume they entered bad data for Town and perform a postcode lookup. If we find address data, then set `CanonicalTown` from the response, or null if there is no address data.

I would expect to see your `GetAddresses` outside of `if (!model.ManualAddressEntry)`?

kevinofficer (approved):

PR #191, opened 2025-06-05
CRNCC-2634 Refactor duplicated code

paul-tennyson:


BPOR/BPOR.Content/Views/Shared/_Layout.cshtml
```diff
@@ -111,25 +111,23 @@
             if (Cookies.get("cookiesAccepted")) {
                 closeCookieBanner();
             }
-            if (Cookies.get("cookiesAccepted").includes("true")) {
-                cookiesAlreadyPresent();
-            }
-
+      
+            handleAcceptedCookies();
```
con-nihr, line 115: Minor: I would move this up just below [112 ](https://github.com/PA-NIHR-CRN/dte-study-api/pull/191/files#diff-60c18293e92d4c5da2cac33d50623ca28350d4eeda50e5514d0117805885f5e8R112) to keep all the similar behaviour together. Even if we have the extra check in `handleAcceptedCookies()`.

con-nihr (approved):

PR #247, opened 2025-06-11
 CRNCC-2481/PRB0040457 Targeted fix to address immediate issue of participant records with missing values.

kira-clarke:


paul-tennyson (approved):

BPOR/BPOR.Rms/Services/CampaignService.cs
```diff
@@ -222,6 +222,13 @@ private async Task QueueNotificationsAsync(List<CampaignParticipantDetails> volu
     {
         foreach (var volunteer in volunteers)
         {
+            var validity = ValidateParticipantForCampaignType(volunteer, campaign.TypeId);
+            if (!validity.isValid)
+            {
+                logger.LogError("Participant ID:{ParticipantId} is not a valid notification target becuase {Reason}", volunteer.Id, validity.failureMessage);
```
con-nihr, line 228: Typo: becuase -> because

BPOR/BPOR.Rms/Services/CampaignService.cs
```diff
@@ -312,6 +313,37 @@ private async Task QueueNotificationsAsync(List<CampaignParticipantDetails> volu
 
         await notificationContext.SaveChangesAsync(cancellationToken);
     }
+
+    private (bool isValid, string? failureMessage) ValidateParticipantForCampaignType(CampaignParticipantDetails volunteer, ContactMethodId campaignTypeId)
```
con-nihr, line 317: It would be good if this could return a list, as there may be more than one reason for a failure.

This might be an opportunity to apply the `IValidatableObject` pattern.

https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.ivalidatableobject.validate?view=net-9.0#system-componentmodel-dataannotations-ivalidatableobject-validate(system-componentmodel-dataannotations-validationcontext)

con-nihr (changes requested):
Minor fix for the typo. Consider returning a list of errors for subsequent releases.

Can you target this PR at `rc/BPOR/v1.40.0`?

https://github.com/PA-NIHR-CRN/dte-study-api/tree/rc/bpor/v1.40.0

con-nihr (approved):

PR #249, opened 2025-06-18
add meta data, image and alt coming from contentful

kevinofficer:


BPOR/BPOR.Content/CampaignPage.cs
```diff
@@ -8,6 +8,10 @@ public class CampaignPage
 
         public string Description { get; set; }
         public string Keywords { get; set; }
+        public Asset MetaImage { get; set; }
```
con-nihr, line 11: This is good work, though this will require a content model update in Contentful too.

For this quick 'patch' I would like to limit what we need to change to release this.

Can we go with just the full embedded HTML for the social media cards in the `head`? (I acknowledge this might be the wrong choice if they decide to change the image again.)

The content managed functionality can be rolled into the next update.

Also for the future enhancement...
Given that there are quite a few metadata properties at the top level of the campaign page I think it would be worth considering wrapping them all up in a child `PageMetadata` content type for grouping and reuse.

con-nihr (changes requested):

kevinofficer:
ticket cancelled

PR #259, opened 2025-06-26
Rms/improvement/crncc 2642 query batching

kira-clarke:
Added batching support for campaign volunteer filtering.


BPOR/BPOR.Rms.Tests/StatsUtils.cs
```diff
@@ -0,0 +1,130 @@
...
+            double sum = 0.0;
+            for (int i = 0; i < observed.Length; ++i)
+            {
+                sum += ((observed[i] - expected[i]) *
+                  (observed[i] - expected[i])) / expected[i];
+            }
+            return sum;
+        }
+       
+        public static double ChiSqPval(double x, int df)
```
con-nihr, line 16: Nicely done with this. However I'm a little nervous rolling our own stats functions.

What are your thoughts on using `MathNet.Numerics.Distributions.ChiSquared.CDF(df, x)` instead?

BPOR/BPOR.Rms/Controllers/FilterController.cs
```diff
@@ -64,18 +64,25 @@ public async Task<IActionResult> Index(VolunteerFilterViewModel model, string? a
...
 
+        model.VolunteerCount = 0;
+        model.Testing.VolunteerResults = Page<VolunteerResult>.Empty();
+
         if (activity == "FilterVolunteers")
         {
-            results = await FilterVolunteersAsync(model, cancellationToken);
+            if (ModelState.IsValid)
+            {
+                model.VolunteerCount = await volunteerFilterService.GetFilteredVolunteerCountAsync(model, cancellationToken);
```
con-nihr, line 74: These were originally deferred queries so we get multiple result sets (count + testing results, if requested) in one network roundtrip to the db.

Is that pattern causing issues?

con-nihr (commented):
Some initial thoughts

BPOR/BPOR.Rms/Controllers/FilterController.cs
kira-clarke: The count part was timing out, and this improvement entailed splitting the count into multiple parts to avoid that timeout - whilst it would have been possible to do this while maintaining the deferred execution model (which would necessitate modifying the test result query to operate in batches), I felt that it was lower risk (for a negligible increase in end-to-end execution time) to split this into two methods each with a single responsibility that could be tested independently.

BPOR/BPOR.Rms/Services/VolunteerFilterService.cs
```diff
@@ -0,0 +1,190 @@
...
+            double pageSize = options.Value.InitialPageSize;
+
+            while (pageMin < 1.0 && (!targetGroupSize.HasValue || targetGroupSize.Value > result.Count))
+            {
+
+                double pageMax = pageMin + pageSize;
+                var volunteerQuery = context
+                    .GetRandomSampleOfParticipants(seed, pageMin, pageMax)
+                    .FilterVolunteers(timeProvider, filter)
+                    .Where(v => !string.IsNullOrEmpty(v.Email)); // TODO: This is suspect? Why are we always excluding participants with no email address?
```
kevinofficer, line 111: can we get an answer to this `TODO`. If we are excluding participants because we only want those who are contactable by email then lets make that more clear. 
e.g. a new parameter `isContactableByEmail` or similar.
```
if (isContactableByEmail)
    {
        volunteerQuery = volunteerQuery.Where(v => !string.IsNullOrEmpty(v.Email));
    }
```
How might this be excluding offline registered participants that may not have an email?

kevinofficer (changes requested):

BPOR/BPOR.Rms/Services/VolunteerFilterService.cs
kira-clarke: I added this TODO in response to the offline journey functionality so that it wouldn't get lost. I think it's out of scope for this issue, but does need investigated and addresses as part of an offline journey ticket.

kevinofficer: agreed out of scope, but lets make sure this is tracked somewhere, as a new problem, or added into an existing problem if it makes sense. I recall seeing problem tickets around filtering numbers being inaccurate.

kevinofficer (approved):

PR #260, opened 2025-06-26
Rms/improvement/crncc 2641 optimise campaign filter

kira-clarke:
Added OSGB grid ref to ParticipantLocation and use that data in volunteer filtering.


BPOR/BPOR.Domain/Entities/ParticipantLocation.cs
```diff
@@ -1,6 +1,9 @@
 using BPOR.Domain.Entities.Configuration;
+using Contentful.Core.Models;
 using NetTopologySuite.Geometries;
+using NIHR.Geometry;
```
con-nihr, line 4: Is this polluting the entity model with logic that is too specific to location geometry?

BPOR/BPOR.Domain/Entities/ParticipantLocation.cs
```diff
@@ -14,10 +17,63 @@ public class ParticipantLocation : ISoftDelete, ITimestamped, IPersonalInformati
...
+    public int Northing { get; set; }
     public void Anonymise()
     {
-        Location = Point.Empty;
-        Location.SRID = ParticipantLocationConfiguration.LocationSrid;
+        SetLocation(Point.Empty);
         IsApproximate = true;
     }
+
+    public static ParticipantLocation FromCoordinates(CoordinatesModel coordinates)
```
con-nihr, line 34: Are you steering the calling code towards using a controlled mechanism for setting the location (and associated properties) consistently?

Can some of these methods be `protected` or `private` to make that clearer to the consuming developer?

BPOR/BPOR.Rms/Controllers/FilterController.cs
```diff
@@ -64,18 +66,22 @@ public async Task<IActionResult> Index(VolunteerFilterViewModel model, string? a
...
         }
         else if (activity == "ClearFilters")
         {
             model = ClearFilters(model);
         }
 
-        model.VolunteerCount = results.Count?.Value;
-        model.Testing.VolunteerResults = results.Items?.Value ?? Page<VolunteerResult>.Empty();
-
+        model.VolunteerCount = results.Count?.HasValue == true ? await results.Count.ValueAsync() : default;
```
con-nihr, line 81: This is a debugging convenience or do you want to leave it in the shipped code?

Please pass `cancellationToken` to the `Async()` methods when it is accepted.

BPOR/BPOR.Rms/Controllers/FilterController.cs
```diff
@@ -64,18 +66,22 @@ public async Task<IActionResult> Index(VolunteerFilterViewModel model, string? a
...
+        model.Testing.VolunteerResults = results.Items != null ? await results.Items.ValueAsync() : Page<VolunteerResult>.Empty();
+        sw.Stop();
+        var i = sw.ElapsedMilliseconds;
```
con-nihr, line 84: We're not doing anything with this value, so it should be removed. `i` is a useful name to have available elsewhere in the scope.

If we want to time specific groups of operations I could see a useful extension method of `ILogger<T>` that returns an `IDisposable` wrapping a stopwatch.

e.g.
```
using(_logger.TimeOperation("Filtering")){
    ... // Perform filtering
}
```

The stopwatch is started when the method is called, stops when the scope is disposed and writes the appropriate message to the logs.

This could be added to the `NIHR.Infrastructure` package.

con-nihr (changes requested):
Great work. Some initial thoughts.

BPOR/BPOR.Rms/Controllers/FilterController.cs
kira-clarke: Cancellation token duly passed.

I don't understand what you are referring to as a 'debugging convenience'

BPOR/BPOR.Rms/Controllers/FilterController.cs
kira-clarke: That should have been removed - apologies - and, yes, having a disposable logging timer is a pattern I've used extensively and is on my list of things for an SDK!

BPOR/BPOR.Domain/Entities/ParticipantLocation.cs
```diff
@@ -14,10 +17,63 @@ public class ParticipantLocation : ISoftDelete, ITimestamped, IPersonalInformati
...
+    {
+        Location = location;
+        SetOsgbFromLocation();
+    }
+
+    public void SetOsgbFromLocation()
+    {
+        if (Location == Point.Empty)
+        {
+            Easting = 0;
```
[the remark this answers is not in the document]
kira-clarke: Setting these two values to zero gives a location in the middle of sea - it's not great (in that semantically it doesn't reflect what the record means), but I agree that anonymising a location record in the current manner has little value.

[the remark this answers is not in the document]
kevinofficer: would it become useful if we used an approximate location like you mentioned?
e.g. SW1 vs SW1A 1AA or add randomised offsets to coordinates.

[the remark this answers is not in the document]
kira-clarke: The postcode forms part of the participant record and isn't used in geolocation searches (so could probably be randomised or removed) - but we could certainly round the lat/long and OSGB coordinates to the nearest 1/10 degree or 10km (for example). I think this is out of scope for this ticket and needs to be put to the client.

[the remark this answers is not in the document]
kevinofficer: No don't worry about that. Let's just confirm with whoever on the client side if needed. We may need to reassure them that we can insert the data for test cases if needed

kevinofficer (approved):

PR #263, opened 2025-07-01
Merge CRNCC-2645 into Rework-Added-RMS

kira-clarke:
https://nihr.atlassian.net/browse/CRNCC-2645


BPOR/BPOR.Rms/Views/Filter/Partials/_FilterButtonsAndCount.cshtml
```diff
@@ -34,12 +34,12 @@
 
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-one-half govuk-!-margin-top-3">
-                <button type="submit" value="FilterVolunteers" name="activity" class="govuk-button">
+                <button class="govuk-button" asp-action="FilterVolunteers" formmethod="post">
```
con-nihr, line 37: FYI. I believe GET here was a deliberate design decision so that the filter URL could be shareable.

I'll approve to unblock this problem and we can revisit the underlying issue.

PR #272, opened 2025-08-04
Feature/crncc 2691 style package

kevinofficer:


BPOR/NIHR.GovUk.AspNetCore.Mvc/TagHelpers/SkipLinkTagHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+                    .Split(' ', StringSplitOptions.RemoveEmptyEntries)
+                    .ToList() ?? new List<string>();
+
+                if (classList.Contains("skip-link", StringComparer.OrdinalIgnoreCase))
+                {
+                    output.Attributes.SetAttribute("class", "govuk-skip-link");
+
+                    if (!output.Attributes.ContainsName("href"))
+                    {
+                        output.Attributes.SetAttribute("href", "#main-content");
```
con-nihr, line 22: I'm not 100% convinced we want to have default handling of `href` and `content`. The view should probably still be responsible for that, but I like that you've considered this so we align with the GDS templates.

PR #357, opened 2026-01-08
P/crncc 2925 dob timestamp

kevinofficer:
https://nihr.atlassian.net/browse/CRNCC-2925
- Deployment file change, to use S3 bucket storage to overcome a storage size error during deployment.
- Normalise date of birth fields because different timestamps was causing no match found, and a new acccount created.


Tkirkwood (approved):
can you also update the sonarqube workflow, it should be straight forward replacement of the actions and likely a update of the credentials as well

PR #361, opened 2026-01-20
create state and nonce for nhs login

kevinofficer:
create and validate nonce

https://nihr.atlassian.net/browse/CRNCC-2928

### DIFF
--- a/src/Infrastructure/Auth/NhsLoginStateStore.cs
+++ b/src/Infrastructure/Auth/NhsLoginStateStore.cs
@@ -1,0 +1,39 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Application.Contracts;
+using Microsoft.Extensions.Caching.Distributed;
+
+public class NhsLoginStateStore : INhsLoginStateStore
+{
+    private readonly IDistributedCache _cache;
+
+    public NhsLoginStateStore(IDistributedCache cache)
+    {
+        _cache = cache;
+    }
+
+    public async Task StoreAsync(string state, string nonce, CancellationToken ct)
+    {
+        await _cache.SetStringAsync(
+            $"nhslogin:{state}",
+            nonce,
+            new DistributedCacheEntryOptions
+            {
+                AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(10),
+            },
+            ct
+        );
+    }
+
+    public async Task<string?> GetAndDeleteAsync(string state, CancellationToken ct)
+    {
+        var key = $"nhslogin:{state}";
+        var nonce = await _cache.GetStringAsync(key, ct);
+
+        if (nonce != null)
+            await _cache.RemoveAsync(key, ct);
+
+        return nonce;
+    }
+}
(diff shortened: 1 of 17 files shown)


src/Infrastructure/Clients/NhsLoginHttpClient.cs
```diff
@@ -67,6 +84,33 @@ public async Task<NhsUserInfo> GetUserInfoAsync(string accessToken, Cancellation
...
+            if (!jwt.Payload.TryGetValue("nonce", out var nonceObj))
+                throw new AuthenticationException("Nonce missing from id_token");
+
+            var actualNonce = nonceObj?.ToString();
+
+            if (!FixedTimeEquals(actualNonce, expectedNonce))
+                throw new AuthenticationException("Invalid nonce");
+        }
+
+        private static bool FixedTimeEquals(string? a, string? b)
```
con-nihr, line 102: 😒
AI support is fine, but you need to understand all the code it is generating and 'sign off' on it yourself.

Including random methods to reimplement standard library functionality is not acceptable.


There's a pretty well tested and optimised implementation of `String.Equals` in the dotnet runtime.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs#L621

src/StudyApi/Startup.cs
```diff
@@ -65,11 +66,15 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton(emailSettings);
             services.AddSingleton(appSettings);
             services.AddSingleton(contentfulSettings);
-
+            
+            services.AddDistributedMemoryCache();
```
con-nihr, line 70: The StudyAPI runs as a Lambda which does not guarantee shared memory between requests. This cache could disappear at any time.

You would have to setup an external cache for this to work correctly.

src/Infrastructure/Auth/NhsLoginStateStore.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public NhsLoginStateStore(IDistributedCache cache)
+    {
+        _cache = cache;
+    }
+
+    public async Task StoreAsync(string state, string nonce, CancellationToken ct)
+    {
+        await _cache.SetStringAsync(
+            $"nhslogin:{state}",
```
con-nihr, line 19: State and nonce are independent, and should be associated with the current user session, we shouldn't really be using state as the key for nonce.

con-nihr (changes requested):
We need to take a different approach with this as server side caching of the state and nonce parameters is not the right way to go both from a security perspective (must be associated with the user session) and architecturally (no existing shared cache when running as a lambda).

This ticket can be split into two parts.

1. Send the nonce. This is all that is required to meet the NHS login requirement, we don't have to verify it - though that is good practice of course.
2. Verify the nonce. Generate the values for state and nonce client side and store in session storage. Check the callback state against the stored value, pass the stored nonce to the backend (existing nhslogin endpoint) and verify the nonce in the JWT received from the Token endpoint.

kevinofficer:
binning as its not a current priority, and no deadline to introduce this anymore.

PR #369, opened 2026-03-02
NHS Login duplicate participants

kevinofficer:
When converting from BPOR login to NHS Login, a duplicate participant is created in the RMS reporting db.
https://nihr.atlassian.net/browse/CRNCC-2925

- update workflows
- origin view
- updated nhs mapper
- updated MapIdentifiers to protect some values overriding
- fallback match by email and dob


BPOR/BPOR.Domain/Entities/ParticipantIdentifier.cs
```diff
@@ -20,9 +20,6 @@ public ParticipantIdentifier()
 
     public int IdentifierTypeId { get; set; }
     public bool IsDeleted { get; set; }
-    public DateTime CreatedAt { get; set; }
```
con-nihr, line 23: This is (should be) a build failure. `ParticipantIdentifier` can't implement `ITimestamped` without these two properties.

BPOR/BPOR.Domain/Migrations/20260303084349_CRNCC-2596-Participant-origin-view.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Dynamo.Stream.Handler.Migrations
+{
+    /// <inheritdoc />
+    public partial class CRNCC2596Participantoriginview : Migration
```
con-nihr: There is a duplicate class name in another migration here...
https://github.com/PA-NIHR-CRN/dte-study-api/blob/p/CRNCC-2925/BPOR/BPOR.Domain/Migrations/20250925140402_CRNCC-2596-Participant-origin-view.cs#L8

These class definitions will clash and fail to build.

BPOR/BPOR.Domain/Migrations/20260303084349_CRNCC-2596-Participant-origin-view.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(@"
+                CREATE 
+                    ALGORITHM = UNDEFINED 
+                    DEFINER = `dte-stream-s`@`%` 
+                    SQL SECURITY DEFINER
+                VIEW `dte`.`Participants_Origin` AS
+                SELECT
```
con-nihr: How's the performance of this view?

There are a couple of levels of nesting, ordering, filtering, etc.

BPOR/BPOR.Infrastructure/Mappers/NhsParticipantMapper.cs
```diff
@@ -12,7 +12,7 @@ public static DynamoParticipant MapNewUserFromRequestAndParticipant(this DynamoP
             NhsId = request.NhsId,
             NhsNumber = request.NhsNumber,
             Email = request.Email,
-            ParticipantId = request.NhsId,
+            ParticipantId = participant.ParticipantId,
```
con-nihr, line 15: Was this the source of the original bug?

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
```diff
@@ -131,15 +131,39 @@ private async Task<Participant> InsertAsync(Dictionary<string, AttributeValue> i
     {
         var identifiers = participantMapper.ExtractIdentifiers(image);
 
+        var email = image.TryGetValue("Email", out var emailAttr) ? emailAttr.S : null;
```
con-nihr: We don't need to do this for every record, just those with no matching identifier. You can shift this extraction down to just before it is needed.

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
```diff
@@ -131,15 +131,39 @@ private async Task<Participant> InsertAsync(Dictionary<string, AttributeValue> i
...
+        var dobRaw = image.TryGetValue("DateOfBirth", out var dobAttr) ? dobAttr.S : null;
 
-        var targetParticipant = await participantDbContext.GetParticipantByLinkedIdentifiers(identifiers)
+        email = email?.Trim().ToLowerInvariant();
+
+        DateTime? dob = null;
+        if (!string.IsNullOrWhiteSpace(dobRaw) && DateTime.TryParse(dobRaw, out var parsedDob))
+        {
+            dob = parsedDob.ToUniversalTime().Date;
```
con-nihr: I think assuming a timezone for DOB has been the cause of all sorts of issues, particularly when the NHS data seems to have a different representation than we use internally.

DOB is just a date and shouldn't have a timezone anyway, though sometimes we have to store a time because of database / EF constraints.

Note: Universal time is usually the correct choice for storing datetimes in the db though.

Are we confident that all the records in DynamoDB (and MySQL) have a consistent time component and are all stored in UTC?

Can we parse into a DateOnly and remove the timezone conversion?

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
```diff
@@ -131,15 +131,39 @@ private async Task<Participant> InsertAsync(Dictionary<string, AttributeValue> i
...
+        }
+
+        var targetParticipant = await participantDbContext
+            .GetParticipantByLinkedIdentifiers(identifiers)
             .ForUpdate()
             .SingleOrDefaultAsync(cancellationToken);
 
+        if (targetParticipant == null && !string.IsNullOrWhiteSpace(email) && dob.HasValue)
```
con-nihr: The study API is supposed to do this matching when converting a native account, so something has gone wrong there.

Also, we have to ensure this logic is consistent in both places, ideally not duplicated so that any new requirements are applied correctly.

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
```diff
@@ -167,31 +191,43 @@ private async Task ProcessRemoveAsync(DynamoDBEvent.DynamodbStreamRecord record,
...
+            .RemoveRange(participant.ParticipantIdentifiers);
 
-        participantDbContext.ParticipantIdentifiers.RemoveRange(idsToRemove);
-
-        if (participant.ParticipantIdentifiers.Except(idsToRemove).All(x => x.IsDeleted))
-        {
-            participantDbContext.Participants.Remove(participant);
-        }
+        // Remove participant
+        participantDbContext.Participants.Remove(participant);
```
con-nihr: Bear in mind that multiple dynamo records can be linked to the same de-duplicated participant in the reporting database.

Do we want to remove an active NHS participant when a native participant record is deleted?

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
```diff
@@ -167,31 +191,43 @@ private async Task ProcessRemoveAsync(DynamoDBEvent.DynamodbStreamRecord record,
...
-            .Include(x => x.ParticipantIdentifiers)
-            .Include(x => x.ContactMethodId)
-            .SingleOrDefaultAsync(cancellationToken);
+        
+        logger.LogInformation(
+            "ProcessRemoveAsync - identifier values: {Values}",
+            string.Join(", ", identifiers.Select(id => id.Value))
+        );
+
+            var participants = await participantDbContext
```
con-nihr: Do we need to do similar email / dob matching here too?

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs
```diff
@@ -135,13 +136,19 @@ public async Task<Participant> Map(Dictionary<string, AttributeValue> record, Pa
         destination.EthnicGroup = source.EthnicGroup;
         destination.EthnicBackground = source.EthnicBackground;
         destination.GenderIsSameAsSexRegisteredAtBirth = source.GenderIsSameAsSexRegisteredAtBirth;
-        destination.NHSNumber = source.NhsNumber;
+        if (
+            !string.IsNullOrWhiteSpace(source.NhsNumber)
+            && destination.NHSNumber != source.NhsNumber
```
con-nihr: What is the benefit of `destination.NHSNumber != source.NhsNumber`?

If the values are already equal assigning the value has no effect anyway.

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs
```diff
@@ -135,13 +136,19 @@ public async Task<Participant> Map(Dictionary<string, AttributeValue> record, Pa
...
+        {
+            destination.NHSNumber = source.NhsNumber;
+        }
         destination.RemovalOfConsentRegistrationAtUtc = source.RemovalOfConsentRegistrationAtUtc;
         destination.HasLongTermCondition = source.Disability;
         destination.GenderId = _refDataService.GetGenderId(source.SexRegisteredAtBirth);
         destination.CommunicationLanguageId = _refDataService.GetCommunicationLanguageId(source.SelectedLocale);
         destination.DailyLifeImpactId = _refDataService.GetDailyLifeImpactId(source.DisabilityDescription);
-        destination.CreatedAt = source.CreatedAtUtc;
+        destination.CreatedAt = destination.Id == 0 ? source.CreatedAtUtc : destination.CreatedAt;
```
con-nihr, line 167: Originally we wanted to replicate the original creation date from the DynamoDB record. Is that no getting passed through for new items?

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs
```diff
@@ -209,18 +216,29 @@ public async Task<Participant> Map(Dictionary<string, AttributeValue> record, Pa
...
+                {
+                    int typeId = _refDataService.GetIdentifierTypeId(keyName);
+
+                    _logger.LogInformation("ExtractIdentifiers - Found {KeyName} = {Value}", keyName, attrValue.S);
+
+                    identifiers.Add(new Identifier(typeId, guid));
+                }
+                else
+                {
+                    _logger.LogWarning("ExtractIdentifiers - {KeyName} was not a valid GUID: {Value}", keyName, attrValue.S);
```
con-nihr: Is this an issue we have encountered?

Should this be upgraded to an error?

con-nihr (changes requested):
This branch does not build, so I have to assume it is untested.

There are some significant changes in how the volunteer data will be synced up between Dynamo and the reporting database in here, so we need to make sure it is behaving correctly before releasing to production.


```
BPOR\BPOR.Domain\Entities\ParticipantIdentifier.cs(7,51,7,63): error CS0535: 'ParticipantIdentifier' does not implement interface member 'ITimestamped.CreatedAt'
BPOR\BPOR.Domain\Entities\ParticipantIdentifier.cs(7,51,7,63): error CS0535: 'ParticipantIdentifier' does not implement interface member 'ITimestamped.UpdatedAt'

BPOR\BPOR.Domain\Migrations\20260303084349_CRNCC-2596-Participant-origin-view.cs(11,33,11,35): error CS0111: Type 'CRNCC2596Participantoriginview' already defines a member called 'Up' with the same parameter types
BPOR\BPOR.Domain\Migrations\20260303084349_CRNCC-2596-Participant-origin-view.cs(59,33,59,37): error CS0111: Type 'CRNCC2596Participantoriginview' already defines a member called 'Down' with the same parameter types
```

BPOR/BPOR.Infrastructure/Mappers/NhsParticipantMapper.cs
kevinofficer: Yes, the reason why a duplicate participant was created in the RMS reporting db

BPOR/BPOR.Domain/Entities/ParticipantIdentifier.cs
kevinofficer: removed

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
kevinofficer: I don't think so. 
Email and dob matching fallback was created for the Insert because the `identifiers` in `.GetParticipantByLinkedIdentifiers(identifiers)` was the NHS Number (guid stored as the dynamo participant id), and seeing as we didn't have that stored for them already, a duplicate account was created.

When removing, we have the identifiers and they exist already

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs
kevinofficer: removed

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs
kevinofficer: newly created Dynamo records where overwriting this value with their `CreatedAtUtc`, e.g. converting from bpor to nhs

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Mappers/ParticipantMapper.cs
kevinofficer: No we didn't experience it, I was testing for identifiers being incorrect, e.g. the nhs number vs the nhs guid number. Removed

BPOR/BPOR.Domain/Migrations/20260331123754_CRNCC-2596-Participant-origin-view.cs
```diff
@@ -17,43 +17,36 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+                JOIN dte.Participants p 
+                    ON pi.ParticipantId = p.Id
 
-                ) x
-                WHERE x.rn = 1;
+                JOIN dte.SysRefIdentifierType it 
+                    ON pi.IdentifierTypeId = it.Id
+
+                WHERE pi.IsDeleted = 0
+                AND p.IsDeleted = 0;
```
kevinofficer, line 50: Updated to return lowest Id

BPOR/BPOR.Domain/Migrations/20260303084349_CRNCC-2596-Participant-origin-view.cs
kevinofficer: removed dupe, not applied to the db

BPOR/BPOR.Domain/Migrations/20260303084349_CRNCC-2596-Participant-origin-view.cs
kevinofficer: updated to return lowest Id

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
kevinofficer: updated to do later

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
kevinofficer: I might need more info from you here. 
`ProcessRemoveAsync` only runs when a Dynamodb record is deleted, either manually or when someone closes their account. Dynamodb records should never be deleted manually, and closing your account should delete all related data

BPOR/BPOR.Domain/Entities/ParticipantIdentifier.cs
con-nihr: Don't we want this table to have a timestamp though?

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
con-nihr: The data model in DynamoDB has gone through a number of iterations and it is possible that we might want to delete records.

It would be unexpected behaviour if deleting an old dynamoDB record actually deleted a participant if there was still another active record against it in the reporting db.

con-nihr (commented):
Can you review the Sonar warnings too please?

BPOR/BPOR.Domain/Entities/ParticipantIdentifier.cs
kevinofficer: Timestamp was added in Pauls first iteration, although I don't think its required. a `CreatedAt` timestamp would tell us when the `ParticipantIdentifier` was created which could be handy if we wanted to find out when a user was converted or something similar, but we could use the data in Dynamo in these cases instead. I'm not sure what other benefits there would be to having a `CreatedAt`, as we don't use it to query which `ParticipantIdentifier` is the origin, although we could, but I don't believe it would be any better than what we are doing in this PR.
I've correctly rolled back the migrations and created a new one for the view.

BPOR/BPOR.Registration.Stream.Handler/src/BPOR.Registration.Stream.Handler/Handlers/StreamHandler.cs
kevinofficer: what is the expected outcome for this then?
`ExtractIdentifiers()` gets `ParticipantId` and `NhsId`
then `GetParticipantByLinkedIdentifiers()` will return `SingleOrDefaultAsync()` for matching `ParticipantId` or `NhsId`.
then the previous implementation removed all `ParticipantIdentifier`'s for that participant.

I think the previous implementation and this one both have the same undesired outcome you've described?

con-nihr: Yes, there may have been a similar issue with the previous implementation.

You are closer to the sequence of bugs responsible for the inconsistencies we are trying to resolve so I'll trust you to test the scenarios where we have a native VS account, then a NHS account.

If you delete the NHS account record _or_ the VS account record the participant record should remain until both linked source records are removed.

con-nihr (approved):

con-nihr (approved):

PR #379, opened 2026-06-03
CRNCC-3078 Pre-screener emails frontend and database

paul-tennyson:


BPOR/BPOR.Rms/Constants/ResearcherEmails.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace BPOR.Rms.Constants;
+
+public class ResearcherEmails
+{
+    public ResearcherEmails()
+    {
+    }
+
```
AdamChambersPA: Do we need this empty constructor?

BPOR/BPOR.Rms/Constants/ResearcherEmails.cs
```diff
@@ -0,0 +1,22 @@
...
+    private static List<Dictionary<string, string>> ResearcherEmailOptions = new List<Dictionary<string, string>>
+    {
+
+        new Dictionary<string, string> { { "label", "Send introductory email" }, { "value", "1" } },
+        new Dictionary<string, string> { { "label", "Next steps, offer pre-screener" }, { "value", "2" } },
+        new Dictionary<string, string> { { "label", "Next steps, without pre-screener" }, { "value", "3" } }
+
+    };
+
+    public static List<Dictionary<string, string>> getResearcherEmailOptions()
```
AdamChambersPA: Could we rename this method to be PascalCase `GetResearcherEmailOptions()`

BPOR/BPOR.Rms/Controllers/ResearcherEmailController.cs
```diff
@@ -0,0 +1,87 @@
...
+using BPOR.Rms.Models;
+using BPOR.Rms.Models.ResearcherEmail;
+using BPOR.Rms.Startup;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+
+namespace BPOR.Rms.Controllers;
+
+public class ResearcherEmailController(ParticipantDbContext context,
+    ILogger<StudyController> logger) : Controller
```
AdamChambersPA: This logger is typed as `ILogger<StudyController>`, but this class is `ResearcherEmailController`. 

Should this be `ILogger<ResearcherEmailController>` so log entries are categorised under the correct controller?

BPOR/BPOR.Rms/Controllers/ResearcherEmailController.cs
```diff
@@ -0,0 +1,87 @@
...
+        return View(model);
+    }
+
+    [HttpPost]
+    public async Task<IActionResult> SendEmail(ResearcherEmailViewModel model, CancellationToken cancellationToken)
+    {
+        if (Enum.TryParse<ResearcherEmailOptions>(
+                model.SelectedEmailId,
+                out var selectedEmail) &&
+            !model.IsEligibleForPrescreener &&
```
AdamChambersPA: We appear to be trusting `IsEligibleForPrescreener` from the submitted view model when deciding whether `NextStepOfferPreScreener` is allowed.

Since this is client-controlled form data, it could be tampered with.

Could we reload the study eligibility state from the database in SendEmail and validate against that instead?

BPOR/BPOR.Rms/Controllers/ResearcherEmailController.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+
+namespace BPOR.Rms.Controllers;
+
+public class ResearcherEmailController(ParticipantDbContext context,
+    ILogger<StudyController> logger) : Controller
+{
+    // GET
+    public async Task<IActionResult> Index(int studyId, CancellationToken cancellationToken)
```
AdamChambersPA, line 17: Should we be validting user roles on both this and the `POST` method? 

The link is hidden for researchers but you can still access the page directly

BPOR/BPOR.Rms/Views/Study/Partials/Details/_VolunteerTable.cshtml
```diff
@@ -6,9 +6,6 @@
 
 @{
     var isIdentifiable = Model.Study.IsRecruitingIdentifiableParticipants;
-    var action = isIdentifiable ? "UpdateRecruited" : "UpdateAnonymousRecruited";
-    var buttonText = isIdentifiable ? "Add enrolments" : "Update recruitment total";
-    var insetText = isIdentifiable ? "Volunteer enrolments can be added when a campaign has been sent." : "The recruitment total can be updated when a campaign has been sent";
     var isAdmin = CurrentUserProvider.User.HasRole(UserRole.Admin);
     var isResearcher = CurrentUserProvider.User.HasRole(UserRole.Researcher);
```
AdamChambersPA: Looks like all of these variables are unused now, I think they can all be removed

BPOR/BPOR.Rms/Models/ResearcherEmail/ResearcherEmailViewModel.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.ComponentModel.DataAnnotations;
+using BPOR.Rms.Models.PreScreenerEligibility;
+
+namespace BPOR.Rms.Models.ResearcherEmail;
+
+public class ResearcherEmailViewModel
+{
+    public int StudyId { get; set; }
+    [Required(ErrorMessage = "Please select an email to send to the researcher.")]
+    public string? SelectedEmailId { get; set; }
```
AdamChambersPA: Is there a reason why this `SelectedEmailId` is a string here? Could we have it as an `int` which would move some of the validation and parsing out of the controller

AdamChambersPA (changes requested):

BPOR/BPOR.Rms/Controllers/ResearcherEmailController.cs
paul-tennyson: I've updated this to reload the relevant data in the controller method

BPOR/BPOR.Rms/Controllers/ResearcherEmailController.cs
paul-tennyson: Updated to check Admin role in the controller

AdamChambersPA (approved):

PR #382, opened 2026-06-11
CRNCC-3078 RTS API address lookup

paul-tennyson:


BPOR/BPOR.Rms.VolunteerInformation/Controllers/VolunteerStudyInformationController.cs
```diff
@@ -183,14 +185,19 @@ public async Task<IActionResult> SiteSearch(
             ModelState.AddModelError(nameof(SiteSearchModel.SearchTerm),
                 "The postcode you've entered cannot be found.");
         }
-
+
```
kira-clarke: The caching should occur in the underlying service and not in the controller. If you want to be really smart you can create a service decorator to layer caching on top of the service (see https://andrewlock.net/adding-decorated-classes-to-the-asp.net-core-di-container-using-scrutor/)

BPOR/BPOR.Rms.VolunteerInformation/Controllers/VolunteerStudyInformationController.cs
```diff
@@ -183,14 +185,19 @@ public async Task<IActionResult> SiteSearch(
             ModelState.AddModelError(nameof(SiteSearchModel.SearchTerm),
                 "The postcode you've entered cannot be found.");
         }
-
+        
+        cache.Set(
+            $"site-search:{searchTerm}",
```
kira-clarke: The key in a memory cache is an object rather than a string - you can take advantage of this by creating a simple private record type for your key, then you don't have to worry about collisions if another service also uses the same string format.

Use cache.GetOrCreate(...) to avoid multiple API requests for the same postcode.

Normalise the postcode - RTS now contains a type 'Postcode' that can parse and normalise postcodes - You could copy it to RMS (with a TODO comment to move it to the SDK). By normalising the postcode you avoid cache misses if the user repeatedly searches for the same postcode with slightly different formatting.

BPOR/BPOR.Rms.VolunteerInformation/Controllers/VolunteerStudyInformationController.cs
```diff
@@ -183,14 +185,19 @@ public async Task<IActionResult> SiteSearch(
...
+            model.SearchResult,
+            TimeSpan.FromMinutes(10));
```
kira-clarke: This should be a configuration setting with a default value of 10.

BPOR/BPOR.Rms/Startup/DependencyInjection.cs
```diff
@@ -151,7 +152,16 @@ public static IServiceCollection RegisterServices(this IServiceCollection servic
         }
 
         services.AddVolunteerInformation();
-        services.AddSingleton<IRtsAddressSource, TestRtsAddressSource>();
+        
+        var rtsApiSettings = services.GetSectionAndValidate<RtsApiSettings>(configuration);
```
kira-clarke: This whole block should be an extension method in NIHR.Rts.Client.DiExtensions

BPOR/NIHR.Rts.Client/Models/RtsAddress.cs
```diff
@@ -23,7 +25,8 @@ public RtsAddress(int id, string addressLine1, string? addressLine2, string? add
     public string? AddressLine4 { get; set; }
     public string? AddressLine5 { get; set; }
     public string Postcode { get; set; }
-    public int Id { get; set; }
+    [JsonPropertyName("Identifier")]
+    public string Id { get; set; }
```
kira-clarke: We have opportunity now to change the property name to `Identifier` easily :-)

BPOR/NIHR.Rts.Client/RtsAddressSource.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        _tokenService = tokenService;
+        _httpClient = httpClient;
+    }
+
+    public async Task<IEnumerable<RtsAddress>> SearchByPostcode(
+        string postcode,
+        CancellationToken cancellationToken)
+    {
+        var accessToken = await _tokenService.GetAccessTokenAsync();
```
kira-clarke: Pass and use the cancellation token

kira-clarke: Do we really need to get an access token for each and every request?

kira-clarke (changes requested):

BPOR/NIHR.Rts.Client/Settings/RtsApiSettings.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace NIHR.Rts.Client.Settings;
+
+public class RtsApiSettings
+{
+    public static string SectionName => "RtsApiSettings";
+    public string BaseUrl { get; set; }
+    public string TokenUrl { get; set; }
+    public string ClientId { get; set; }
+    public string ClientSecret { get; set; }
+    public string AddressCacheTimeSpanMinutes { get; set; }
```
kira-clarke: This should be an integer! I would also call it 'AddressCacheTtlMinutes' (TTL = Time To Live)

BPOR/NIHR.Rts.Client/Settings/RtsApiSettings.cs
```diff
@@ -0,0 +1,12 @@
...
+    public string TokenCacheTimeSpanHours { get; set; }
```
kira-clarke: Also an integer, and use minutes rather than hours (Access token are short-lived, typically no more than an hout but can be less), and call it TokenCacheTtlMinutes

BPOR/NIHR.Rts.Client/RtsAddressSource.cs
```diff
@@ -0,0 +1,101 @@
...
+        var token = await _tokenService.GetAccessTokenAsync(cancellationToken);
+
+        using var request = new HttpRequestMessage(
+            HttpMethod.Get,
+            requestUri);
+
+        request.Headers.Authorization =
+            new AuthenticationHeaderValue("Bearer", token);
+
+        var response = await _httpClient.SendAsync(
```
kira-clarke, line 99: I don't see how this is retried with a fresh token if it returns 401/403

BPOR/BPOR.Rms.VolunteerInformation/Controllers/VolunteerStudyInformationController.cs
```diff
@@ -176,15 +177,16 @@ public async Task<IActionResult> SiteSearch(
         {
             return View(model);
         }
-
-        model.SearchResult = (await addressSource.SearchByPostcode(model.SearchTerm, cancellationToken)).ToArray();
+        
+        var parsedPostcode = ParsePostcode(model.SearchTerm);
```
kira-clarke: If I enter a value that is not a syntactically correct postcode then I get a 500 error.
 
I would change the parameter to a SiteSearchModel, then take a look at `MaxWordValidator` and use the same pattern to validate the `searchterm` property as a postcode.

BPOR/NIHR.Rts.Client/RtsAddressSource.cs
```diff
@@ -0,0 +1,101 @@
...
+                               Convert.ToDouble(_settings.AddressCacheTimeSpanMinutes));
+
+                       var response = await SendWithTokenRetryAsync(
+                           $"api/v2/Rts/GetOrganisationList?postcode={postcode}",
+                           cancellationToken);
+
+                       var result = await response.Content.ReadFromJsonAsync<RtsResponse>(
+                           cancellationToken: cancellationToken);
+
+                       return result?.Result.RtsOrganisations?.ToArray()
```
kira-clarke: I would cache each individual address as well so that a later call to GetById hits the cache rathern than RTS.

kira-clarke (changes requested):

PR #384, opened 2026-06-11
Merge MS1 into RMS 1.9.0

kira-clarke:


BPOR/BPOR.Domain/Entities/ParticipantDbContextFactory.cs
```diff
@@ -2,30 +2,16 @@
...
 using NIHR.Infrastructure.EntityFrameworkCore;
+using Pomelo.EntityFrameworkCore.MySql.Infrastructure;
 
 namespace BPOR.Domain.Entities;
 
 public class ParticipantDbContextFactory() : IDesignTimeDbContextFactory<ParticipantDbContext>
 {
     public ParticipantDbContext CreateDbContext(string[] args)
     {
-        // TODO: make this more consistent. Base factory in NIHR.Infrastructure.EntityFrameworkCore.
```
kira-clarke, line 12: This method is only used when generating migrations - I removed this block of code and hardcoded the target MySQL version because the generated SQL should not be sensitive to the environment (i.e. config, env vars and deployed databases); it also prevents migrations from being worked upon when the database isn't available.

BPOR/BPOR.Infrastructure/Authentication/ApiKeyAuthenticationhandler.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Text;
+using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace BPOR.Infrastructure.Authentication;
+
+public sealed class ApiKeyAuthenticationHandler(
```
kira-clarke, line 12: Currently a fairly basic API Key auth handler for a specific case - but easily extendable to work with non-fixed keys etc.

BPOR/BPOR.Rms.VolunteerInformation/Controllers/VolunteerController.cs
```diff
@@ -0,0 +1,92 @@
...
+        long participantId)
+    {
+        if (!options.Value.EnableRrvApi)
+        {
+            return NotFound();
+        }
+        
+        return Ok(new GetTestTokenResponse
+        {
+            Token = new Guid(staticCheckValue, 0, 0, BitConverter.GetBytes(participantId)).ToString("N")
```
kira-clarke, line 32: Simple test token generation - final implementation will use RrvTokenGenerator and produce vaguely sane tokens!!

BPOR/BPOR.Rms/Controllers/ResearcherController.cs
```diff
@@ -279,174 +283,73 @@ public IActionResult AddStudySuccess(AddStudySuccessViewModel viewModel)
...
-        }
-
-        // Recruitment End Date
-
-        if (model.RecruitmentEndDate.Day == null)
+        // All 3 values missing...
+        if (value.Day == null && value.Month == null && value.Year == null)
         {
-            ModelState.AddModelError("RecruitmentEndDate.Day", "Recruitment end date (UK) must include a day");
+            ModelState.AddModelError($"{propertyPath}.Day", $"Enter {propertyCaption}");
```
kira-clarke, line 291: Ideally this should be moved into a FluentValidation extension - however because this was fixed as a pre-existing bug found while testing MS1, this is a a deliberately minimal change to correct the existing behaviour.

PR #390, opened 2026-07-03
CRNCC-3149 Conditional accordion display

paul-tennyson:


kira-clarke (changes requested):
The campaign table still needs to be collapsible for researchers - just use a GDS details component.

PR #392, opened 2026-07-06
Ticket/ms2/cncc 3161

kira-clarke:


kevinofficer (commented):
@kira-clarke looks fine, but could you review the sonar errors and warnings?

kira-clarke:
@kevinofficer Done (there's one left, but honestly 'Loop should be simplified by calling Select(study => study.Entity))' makes the loop more complex).

kevinofficer (approved):

PR #408, opened 2026-07-22
CRNCC-3158 Formatting updates

paul-tennyson:


con-nihr (changes requested):
@paul-tennyson we have to maintain backwards compatibility with the campaign page and we want the have base style sheets and components like `nihr-typography.css` and GovUK radio buttons consistent across all apps. We can't make direct modifications there to only suit the needs of the volunteer information pages.

There should be a more specific stylesheet for these particular pages that we can put any overrides in to fix these specific issues.

Briefly looking at the ticket we might have to have a broader conversation about applying various GOVUK standards.

kira-clarke:
> @paul-tennyson we have to maintain backwards compatibility with the campaign page and we want the have base style sheets and components like `nihr-typography.css` and GovUK radio buttons consistent across all apps. We can't make direct modifications there to only suit the needs of the volunteer information pages.
> 
> There should be a more specific stylesheet for these particular pages that we can put any overrides in to fix these specific issues.
> 
> Briefly looking at the ticket we might have to have a broader conversation about applying various GOVUK standards.

This is meant to be an MVP, and the proposed changes only affect internal administrators. We're trying to establish and maintain a consistent set of styles, and introducing yet more overrides for specific elements is going in the opposite direction (not to mention the maintenance overhead of that going forward)

I think that Becky has a design system in mind for RMS in the long-term, but that really needs to be its own ticket (or set of tickets ... or even a dedicated release) and not scoped to a specific set of elements/pages.

As per Chris's comment, I also think we should be pushing back on these changes and deferring them to a wider discussion around RMS/NIHR styling.

PR #412, opened 2026-07-30
Ticket/main/crncc 3199

kevinofficer:
- Resolve DynamoDB mapping, storing native ParticipantId as intended instead of NhsId.
- Enrich DynamoDB with demographic info then save once, rather than saving, then overwriting with demo info.
- Catch Cognito user not found.
- On deletion, loop DynamoDB records by email until 0 remaining, aiming to resolve the existing data issue if a Participant in the same data situation attempts to delete their account.
- Move emails to the last action.

### DIFF
--- a/src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
+++ b/src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
--- a/src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
+++ b/src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
@@ -86,9 +86,9 @@ namespace Infrastructure.Persistence
         public async Task CreateParticipantDetailsAsync(ParticipantDetails entity)
         {
             // TODO pull out this logic into a mapper
-            entity.Pk = ParticipantKey(string.IsNullOrEmpty(entity.ParticipantId)
-                ? entity.NhsId
-                : entity.ParticipantId);
+            entity.Pk = ParticipantKey(string.IsNullOrEmpty(entity.NhsId)
+                ? entity.ParticipantId
+                : entity.NhsId);
             entity.Sk = ParticipantKey();
 
             await _context.SaveAsync(entity, _config);
--- a/src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
+++ b/src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
@@ -86,9 +86,9 @@ namespace Infrastructure.Persistence
         public async Task CreateParticipantDetailsAsync(ParticipantDetails entity)
         {
             // TODO pull out this logic into a mapper
-            entity.Pk = ParticipantKey(string.IsNullOrEmpty(entity.NhsId)
-                ? entity.ParticipantId
-                : entity.NhsId);
+            entity.Pk = ParticipantKey(string.IsNullOrEmpty(entity.ParticipantId)
+                ? entity.NhsId
+                : entity.ParticipantId);
             entity.Sk = ParticipantKey();
 
             await _context.SaveAsync(entity, _config);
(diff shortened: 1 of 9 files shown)


src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
```diff
@@ -77,6 +77,55 @@ public async Task<ParticipantDetails> QueryIndexForParticipantDetailsAsync(strin
...
+
+                    var participantId = indexParticipant.Pk.Replace("PARTICIPANT#", "");
+
+                    var fullParticipant = await GetParticipantDetailsAsync(participantId);
+
+                    if (fullParticipant != null)
+                    {
+                        results.Add(fullParticipant);
+                    }
+                }
```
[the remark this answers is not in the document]
kevinofficer: resolved

src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -225,26 +224,40 @@ public async Task DeleteUserAsync(string participantId)
...
+                .Append(entity)
+                .Where(x => x != null)
+                .GroupBy(x => x.Pk)
+                .Select(x => x.First())
+                .ToList();
+
+            foreach (var participant in participantsToDelete)
+            {
+                await RemoveParticipantDataAsync(participant);
+            }
```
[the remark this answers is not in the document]
kevinofficer: resolved

src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -104,8 +104,8 @@ private async Task CreateUserAndDeactivateOldUserAsync(ParticipantDetails reques
             .GetParticipantDemographicsAsync(participant.Pk.Replace("PARTICIPANT#", ""));
         if (demographics.HasDemographics)
         {
+            entity.ApplyDemographics(demographics);
             await _participantRepository.CreateParticipantDetailsAsync(entity);
```
kira-clarke: `AddDemographicsToNhsUserAsync` is no longer called and can be removed from both the interface and implementation

kira-clarke (commented):
Caveat: I've added a missing using directive that was preventing compilation; I've not had time to execute this code yet but I'm approving this on the basis that it has been tested by Kevin.

I'm satisfied from code inspection that this PR addresses two issues:

1. When a participant is migrated from native to NHSID, a new DynamoDB record is created, which is then overwritten by the old record and the NHSID is not saved.
2. When a participant is deleted, all participant records with the target email address are now anonymised rather than the single record with the matching participant ID

The code has multiple redundant operations and a redundant loop - I've approving this PR on the basis that these need to be addressed when time allows.

src/Infrastructure/Persistence/ParticipantDynamoDbRepository.cs
```diff
@@ -77,6 +77,50 @@ public async Task<ParticipantDetails> QueryIndexForParticipantDetailsAsync(strin
...
+                        results.Add(participant);
+                    }
+                }
+
+                lastEvaluatedKey = response.LastEvaluatedKey?.Count > 0 ? response.LastEvaluatedKey : null;
+            }
+            while (lastEvaluatedKey != null);
+
+            return results
+                .GroupBy(x => x.Pk)
```
kira-clarke: Unless we are expecting multiple records for the same PK, this group, select and to list operation is not needed

src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -225,26 +226,52 @@ public async Task DeleteUserAsync(string participantId)
...
+            }
+
+            var email = entity.Email;
+            var selectedLocale = entity.SelectedLocale ?? SelectedLocale.Default;
+
+            await SaveAnonymisedDemographicParticipantDataAsync(entity);
+
+            var deletedPks = new HashSet<string>(StringComparer.Ordinal);
+
+            while (true)
```
kira-clarke: This loop should not be required, nor should deletedPks.

src/Infrastructure/Services/ParticipantService.cs
```diff
@@ -225,26 +226,52 @@ public async Task DeleteUserAsync(string participantId)
...
+            {
+                var participants = await _participantRepository.GetAllParticipantDetailsByEmailAsync(email);
+
+                var participantsToDelete = participants
```
kira-clarke: This entire statement can be removed

kira-clarke (approved):
Approved as this is a P1 - but see other comments.

kevinofficer:
Changed based after discussing branching and versioning with Chris

PR #414, opened 2026-08-10
CRNCC-2718 - Re-instate fix for PRB0040556: BPoR: Volunteer Stage2Complete value getting overwritten

con-nihr:
CRNCC-2718 - Re-instate fix for PRB0040556: BPoR: Volunteer Stage2Complete value getting overwritten which was removed from the stream handler when v1.9.0 was deployed.


kira-clarke (approved):
Approved as this a reinstatement of a previously approved change.

I did go down a rabbit hole of trying to work out why the three entity reload operations were added - I note that the commit comments state this is to facilitate testing, but I'm still curious.

